PR #71, opened 2020-07-24
Restore suggestion behaviour

jonathonherbert:
_Co-authored-by: @SHession, @tjsilver_ 

## What does this change?

Suggestions were removed in #64. This PR adds them back, and improves UI behaviour –
- When suggestions are applied, the tooltip they came from is closed.
- When the document is changed, matches touched by those changes are removed.

suggestion-towne

## How to test

Launch the plugin and check the behaviours listed above occur for matches that have suggestions.

Examples from the current Guardian sheet include 'lowlives', which gives a single suggestion.

The behaviour above should be reflected in the UI.

### DIFF
--- a/src/ts/commands.ts
+++ b/src/ts/commands.ts
@@ -79,14 +79,30 @@ export const requestMatchesForDirtyRangesCommand = (
  * over a match decoration) to allow the positioning of e.g. tooltips.
  */
 export const indicateHoverCommand = (
-  matchId: string | undefined,
-  hoverInfo: IStateHoverInfo | undefined
+  matchId: string,
+  hoverInfo?: IStateHoverInfo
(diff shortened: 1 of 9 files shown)


src/ts/commands.ts
```diff
@@ -79,8 +79,8 @@ export const requestMatchesForDirtyRangesCommand = (
  * over a match decoration) to allow the positioning of e.g. tooltips.
  */
 export const indicateHoverCommand = (
-  matchId: string | undefined,
-  hoverInfo: IStateHoverInfo | undefined
+  matchId: string,
+  hoverInfo?: IStateHoverInfo
```
jonathonherbert, line 83: This API has been narrowed in favour of a separate function for `stopHover`.

PR #72, opened 2020-07-24
Improve developer experience via docs, npm commands

jonathonherbert:
## What does this change?

Adds additional documentation improves commands to clarify processes for development and release.

## How to test

The commands in the NPM scripts should work as expected.

If the docs give enough guidance for new devs to be able to confidently run the project locally and issue releases.


build/README.md
```diff
@@ -3,8 +3,22 @@ This Prosemirror plugin adds the ability to validate a document by sending it, o
 
 It's still in its early stages! [There's a demo here.](https://guardian.github.io/prosemirror-typerighter/) See the 'pages' directory for an example implementation.
 
-Short-term roadmap:
```
jonathonherbert, line 6: This no longer reflects our roadmap, so best remove it for now.

PR #75, opened 2020-07-29
Add feedback link

aug24:
## What does this change?

Adds a link to the typerighter popup so allow the user to send feedback via a google form.

The google form has been set up to only accept entries from validated email addresses ending @guardian.co.uk, but is currently owned by Justin Rowles and should be owned by an internal account.

## How to test

I... don't really know.  I ran composer locally and installed the plugin locally.

Clickable link with appropriate embedded metadata appears in popup.

Form sends email to typerighter@theguardian.com

## Images

image

### DIFF
--- a/src/ts/components/Match.tsx
+++ b/src/ts/components/Match.tsx
@@ -41,7 +41,7 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
             <div className="MatchWidget__feedbackLink">
               <a
                 target="_blank"
-                href={this.getFeedbackLink(feedbackInfo)}
+                href={this.getFeedbackLink(this.props.feedbackHref!, feedbackInfo)}
               >
                 Something's not right? Tell us!
               </a>
@@ -52,9 +52,9 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
     );
   }
 
-  private getFeedbackLink = (feedbackInfo: any) => {
+  private getFeedbackLink = (feedbackHref: string, feedbackInfo: any) => {
     const data = encodeURIComponent(JSON.stringify(feedbackInfo, undefined, 2))
-    return this.props.feedbackHref! + data
+    return feedbackHref + data
   }
 
 }
--- a/src/ts/components/Match.tsx
+++ b/src/ts/components/Match.tsx
@@ -8,6 +8,11 @@ interface IMatchProps<TMatch extends IMatch> {
   match: TMatch;
 }
 
+const getFeedbackLink = (feedbackInfo: any) => {
+  const data = encodeURIComponent(JSON.stringify(feedbackInfo, undefined, 2))
+  return "https://docs.google.com/forms/d/e/1FAIpQLSfMOgvJtCchnW0_2zB7Afz_WtYJ5lnPqQI-dgFZ-p0B4h6uKw/viewform?usp=pp_url&entry.110962249=" + data
+}
+
 class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
   public ref: HTMLDivElement | null = null;
   public render({
@@ -15,6 +20,7 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
     applySuggestions
   }: IMatchProps<TMatch>) {
     const suggestionsToRender = replacement ? [replacement] : suggestions || [];
+    const feedbackInfo = { matchId, category, message, suggestions, replacement };
     return (
       <div className="MatchWidget__container">
         <div className="MatchWidget" ref={_ => (this.ref = _)}>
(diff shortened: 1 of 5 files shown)


src/ts/components/Match.tsx
```diff
@@ -32,6 +38,14 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
                 matchId={matchId}
                 suggestions={suggestionsToRender}
               />
+              <div className="MatchWidget__feedbackLink">
+                <a
```
SHession, line 42: This div is within `MatchWidget__suggestion-list` which is not meant to display when there are no suggestions. I recommend you move it out.

src/ts/components/Match.tsx
```diff
@@ -8,13 +8,19 @@ interface IMatchProps<TMatch extends IMatch> {
   match: TMatch;
 }
 
+const getFeedbackLink = (feedbackInfo: any) => {
+  const data = encodeURIComponent(JSON.stringify(feedbackInfo, undefined, 2))
+  return "https://docs.google.com/forms/d/e/1FAIpQLSfMOgvJtCchnW0_2zB7Afz_WtYJ5lnPqQI-dgFZ-p0B4h6uKw/viewform?usp=pp_url&entry.110962249=" + data
```
SHession, line 13: `nit` Is there any reason this function is not part of the component class?

src/ts/components/Match.tsx
```diff
@@ -32,6 +38,14 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
                 matchId={matchId}
                 suggestions={suggestionsToRender}
               />
+              <div className="MatchWidget__feedbackLink">
+                <a 
+                  target="_blank"
+                  href={getFeedbackLink(feedbackInfo)}
+                >
+                  Something's not right? Tell us!
```
SHession, line 46: `nit` I wonder if there is a nice way to display this message, possibly an icon?

src/ts/components/Match.tsx
```diff
@@ -8,13 +8,19 @@ interface IMatchProps<TMatch extends IMatch> {
...
+}
+
 class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
   public ref: HTMLDivElement | null = null;
   public render({
     match: { matchId, category, message, suggestions, replacement },
     applySuggestions
   }: IMatchProps<TMatch>) {
     const suggestionsToRender = replacement ? [replacement] : suggestions || [];
+    const feedbackInfo = { matchId, category, message, suggestions, replacement };
```
SHession, line 23: I don't think this includes the original match text. Which could be helpful in determining if a match has been mistakenly made.

src/ts/components/Match.tsx
aug24: There's a suggested icon in the mockup, but also this exact text.  I haven't done the icon ... yet.

tjsilver, line 46: `nit` Not sure if this reads very well - maybe it could change to 'Something not right? Tell us!' Or do we need to be even more specific that something's not right with **this particular** match?

jennygrahamjones, line 46: Maybe?

```suggestion
                  Something not right with this suggestion? Tell us!
```

src/ts/components/Match.tsx
jennygrahamjones, line 13: Could the form URL be provided as a config value?

aug24: Just doing that :)

src/ts/components/Match.tsx
aug24: This text is from the UI design.  I'm not changing it.

aug24: This text is from the UI design.  I'm not changing it.

src/ts/components/Match.tsx
aug24: Yes - that's too hard for me for now.  I'm going to do that as stage 2.

src/ts/components/Match.tsx
aug24: Done.

src/ts/components/Match.tsx
```diff
@@ -34,10 +37,26 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
...
+            </div>
+          )}
         </div>
       </div>
     );
   }
+
+  private getFeedbackLink = (feedbackInfo: any) => {
+    const data = encodeURIComponent(JSON.stringify(feedbackInfo, undefined, 2))
+    return this.props.feedbackHref! + data
```
jennygrahamjones, line 57: Could we pass `feedbackHref` as an argument to this function, rather than referencing the prop? It seems like we'd only want to call this when we definitely have the href.

aug24: Done.  Although the logic made that implicit (only called when feedbackHref exists) that makes it explicit.

src/ts/createView.tsx
```diff
@@ -18,7 +18,8 @@ const createView = (
   commands: Commands,
   sidebarNode: Element,
   controlsNode: Element,
-  contactHref: string
+  contactHref?: string,
+  feedbackHref?: string
```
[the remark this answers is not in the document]
aug24: On these (three) issues of naming and being more specific about feedback on rules - these are directly taken from the UI design, so I'm unwilling to override the designer.

I have taken the opportunity to make the contactHref (which appears in the sidebar, all the time) optional, as I was doing the same for the feedbackHref.  I'd like user feedback from CP before changing further.

PR #77, opened 2020-08-03
Add error messaging

tjsilver:
## What does this change?

At the moment, when we make requests to the Typerighter API, it's not clear if something has gone wrong or not.

In this PR we add an error state to the UI that reflects any problems, meeting these requirements:

- The error state should describe what's wrong in as clear a way as possible to the user, who will not be a developer.
- The error should clearly be 'owned' by Typerighter, not the parent tool. This probably means colocating it in existing UI.
- The error should be dismissable.

## How to test

To emulate a partial failure on your local environment, you can go to the Typerighter `ApiController.scala` and use `scala.util.random` to reject arbitrary requests.

To emulate a complete failure, block the Typerighter domain requests.

Click 'Check whole document' and the error message should appear below.

A user is able to tell when something has gone wrong and is informed of what actions to take to rectify the situation.

## Images

TyperighterError


src/ts/components/Controls.tsx
```diff
@@ -13,6 +13,7 @@ interface IProps {
   getCurrentCategories: () => ICategory[];
   addCategory: (id: string) => void;
   removeCategory: (id: string) => void;
+  contactHref?: string;
```
jonathonherbert, line 16: To be specific – the error contact href?

src/ts/services/adapters/TyperighterAdapter.ts
```diff
@@ -78,7 +78,8 @@ class TyperighterAdapter implements IMatcherAdapter {
         onRequestError({
           requestId,
           blockId: input.id,
-          message: e.message
+          message: e.message,
+          categoryIds
```
jonathonherbert, line 82: 👍 great spot, this may be handy later!

src/ts/state/reducer.ts
```diff
@@ -122,8 +122,10 @@ export interface IPluginState<TMatches extends IMatch = IMatch> {
   requestsInFlight: {
     [requestId: string]: IBlocksInFlightState;
   };
-  // The current error status.
-  error: string | undefined;
+  // The current error message.
+  errorMessage: string | undefined;
+  // The current error state.
+  hasError: boolean;
```
jonathonherbert, line 128: Interesting ... is it possible to be in an error state without a message (`errorMessage: undefined`)? If not, is `hasError` redundant? (This may be inconsequential re: other changes we're about to make, so it's only a nit!)

jonathonherbert:
Oh, one question – is it possible for the user to dismiss this error? Is it automatically cleared when they run another check?

src/ts/state/reducer.ts
SHession: I have replaced this with a selector per our conversation earlier.

SHession:
> Oh, one question – is it possible for the user to dismiss this error? Is it automatically cleared when they run another check?

That's right, it's cleared in the request start `handleRequestStart`

PR #79, opened 2020-08-04
Numerous styling changes

jonathonherbert:
## What does this change?

The UI for this plugin was looking a little frumpy. It's still a bit frumpy, but I've made a few changes –

- References to category colours are now represented by swatches or borders, to ensure the category text is always legible
- A bug which made it difficult to traverse from match to tooltip has been fixed
- The two panels are consolidated – I'm not sure there was a good reason to keep them separate
- Font sizes have been normalised into three variables to ensure things are consistent, and applied at the root with em sizing to ensure sizing is consistent when this UI is embedded elsewhere 
- A few spacing tweaks, also to keep things consistent
- The suggestion/replacement rendering has been moved to the top of the match tooltip, the bounding box of the suggestion is now clear from a background colour, and the styling of the form link toned down, which I think makes for a better UI.

Before:
tr-before-makeover
After:
tr-makeover

Phew, that's a lot of changes! I'm happy to split into separate PRs – all of the points above were fairly trivial to implement but I appreciate it might not be straightforward to understand what does what.

Re: design, all of this is of course subjective, and comments/crits are welcomed!

## How to test

Take the UI for a spin!

### DIFF
--- a/pages/dist/index.html
+++ b/pages/dist/index.html
@@ -62,10 +62,6 @@
         border-radius: 4px 4px 0 0;
       }
 
-      #controls {
-        flex-shrink: 0;
-      }
-
       #sidebar {
         min-height: 0;
       }
@@ -86,7 +82,6 @@
           </div>
         </div>
         <div class="Sidebar">
-            <div id="controls" class="Sidebar__container"></div>
             <div id="sidebar" class="Sidebar__container"></div>
         </div>
       </div>
--- a/src/css/index.scss
+++ b/src/css/index.scss
@@ -12,10 +12,9 @@ $match-suggestion-background-color-hover: #eee;
 $match-suggestion-color: #999;
 $match-suggestion-color-hover: #888;
 $match-suggestion-wiki-extract-height: 85px;
-$font-size-base: 14px;
+$font-size-base: 15px;
 $font-size-large: 1.1em;
 $font-size-small: 0.8em;
-$font-size-x-small: 0.7em;
 
 // Mixins
 @mixin placeholderAnimation {
@@ -100,7 +99,7 @@ hr {
   font-weight: normal;
   box-shadow: 0 14px 28px rgba(0, 0, 0, 0.2), 0 10px 10px rgba(0, 0, 0, 0.18);
   border-radius: 3px;
-  width: 300px;
+  width: 350px;
   transition: opacity 0.1s, transform 0.1s;
   transform: translate3d(0, -3px, 0);
   z-index: 100;
@@ -180,7 +179,7 @@ hr {
 
 .MatchWidget__feedbackLink {
   margin-top: $gutter-width;
-  font-size: $font-size-x-small;
+  font-size: $font-size-small;
   a {
     color: $match-gray;
   }
--- a/src/css/index.scss
+++ b/src/css/index.scss
@@ -1,6 +1,6 @@
 // Variables
 $gutter-width: 10px;
-$sidebar-border-color: #ccc;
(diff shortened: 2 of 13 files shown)


pages/dist/index.html
```diff
@@ -62,10 +62,6 @@
         border-radius: 4px 4px 0 0;
       }
 
-      #controls {
```
jonathonherbert, line 65: This element is no longer needed, as the controls and results panes have been consolidated.

src/css/index.scss
```diff
@@ -84,14 +91,14 @@ hr {
...
   display: block;
   position: relative;
+  padding: $gutter-width * 1.5;
   font-family: sans-serif;
-  font-size: 14px;
   font-style: normal;
   font-weight: normal;
-  background-color: white;
-  box-shadow: #4e56643b 0px 0px 1px 0px, #3f48573a 0px 2px 7px 0px;
+  box-shadow: 0 14px 28px rgba(0, 0, 0, 0.2), 0 10px 10px rgba(0, 0, 0, 0.18);
```
jonathonherbert, line 101: Stolen from material-ui, which does some tidy shadows.

src/ts/components/MatchOverlay.tsx
```diff
@@ -70,7 +70,7 @@ class MatchOverlay<TMatch extends IMatch = IMatch> extends Component<
             // We hoist top slightly to ensure that the rendered element overlaps the
             // span that triggered the overlay -- if the mouse falls through a gap it
             // will trigger a mouseleave event that will close the overlay.
-            top: top - 1,
+            top: top - 5,
```
jonathonherbert, line 73: This hoists the tooltip more aggressively – there's a countervailing change to the tooltip margin to offset this.

pages/dist/index.html
tjsilver: Consolidating the panes is really helpful, it will make it clearer to users which tool they're in.

SHession: One counter point to having combined planes, is I feel the results are less clearly grouped under the "Results" header. I wonder if this could be subtly achieved by making the results a slightly different colour or an offset?

jonathonherbert: Agreed, 'Results' and 'Typerighter' are the same weight and it's a bit strange. Honestly I'd favour just removing 'Typerighter' altogether for now – testers already know what they're in for – as I'm confident UX changes down the line will establish that context if it's still needed.

SHession: I would be interested to see that, I think it's worth a try.

src/css/index.scss
```diff
@@ -136,31 +155,35 @@ hr {
...
 }
 .MatchWidget__suggestion + .MatchWidget__suggestion {
   margin-top: 3px;
 }
 
 .MatchWidget__feedbackLink {
-  text-align: right;
-  font-size: 12px;
+  margin-top: $gutter-width;
+  font-size: $font-size-x-small;
```
tjsilver, line 183: This font size is so small it's a bit hard to see.

jonathonherbert:
@tjsilver I think this is the result of an odd set of rules in our sheet – I can reproduce, but I don't think it's a problem with this PR! We'll fix ASAP.

src/css/index.scss
jonathonherbert: Agreed – shall we bring it up a px, or perhaps just ignore it in favour of `$font-size-small`?

tjsilver: Yeah maybe the latter - `$font-size-small`?

PR #80, opened 2020-08-05
Add match context to model

aug24:
## What does this change?
Adds a  'matchContext' field to the IMatch type so that we can pass it back out to the feedback form.

## How to test

Install on local compower and run local typerighter. This has been done successfully.

See also: https://github.com/guardian/typerighter/pull/59

Feedback form has matchContext in the technical data section.

No

## Images
N/A


SHession:
This PR is a bit of date and doesn't account for the auth changes made to the Typerighter service. Rebasing master would make it easier to test.

PR #84, opened 2020-08-06
Add logging capability

aug24:
## What does this change?

Adds an optional logging capability to the plugin.

https://trello.com/c/dU27JPiI/317-log-prosemirror-typerighter-errors-to-elk

## How to test
Initial logging should be visible in the javascript console.

See logging.

N/A

## Images
N/A

### DIFF
--- a/src/ts/createView.tsx
+++ b/src/ts/createView.tsx
@@ -7,6 +7,7 @@ import Controls from "./components/Controls";
 import { Commands } from "./commands";
 import { IMatch } from "./interfaces/IMatch";
 import { MatcherService } from ".";
+import { ILogger, consoleLogger } from "./utils/logger"
 
 /**
  * Instantiate a UI for the given EditorView, commands, and configuration,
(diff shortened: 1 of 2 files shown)


src/ts/createView.tsx
```diff
@@ -35,6 +38,7 @@ const createView = (
   view.dom.parentNode!.replaceChild(wrapperElement, view.dom);
   wrapperElement.appendChild(view.dom);
   view.dom.insertAdjacentElement("afterend", overlayNode);
+  logger.info("Typerighter plugin starting");
```
jonathonherbert, line 41: Do we still need this line?

jonathonherbert: Oh, wait, I see what's happening – ignore 🤦

jonathonherbert: (I'd favour removing it as soon as our logger is doing useful work elsewhere)

aug24: Yes,  that's the intent.

PR #85, opened 2020-08-06
Remove tick from correct matches

SHession:
## What does this change?

This PR simply removes the `markAsCorrect` tick from matches where the suggestion perfectly matches the text. Instead it marks those matches as the currently used green colour #3ff200.

## How to test

Check for matches on a phrase where the suggestion matches the word exactly, an example is Instagram. The decoration for this match should no longer include a tick, and should always appear green.

There is less noise on the document when checking for matches.

### DIFF
--- a/src/ts/utils/decoration.ts
+++ b/src/ts/utils/decoration.ts
@@ -5,7 +5,6 @@ import { IRange, IMatch } from "../interfaces/IMatch";
 
 // Our decoration types.
 export const DECORATION_MATCH = "DECORATION_MATCH";
-export const DECORATION_MATCH_IS_CORRECT = "DECORATION_MATCH_IS_CORRECT";
 export const DECORATION_MATCH_IS_SELECTED = "DECORATION_MATCH_IS_HOVERING";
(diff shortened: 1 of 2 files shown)


src/ts/utils/decoration.ts
```diff
@@ -121,10 +107,12 @@ export const createDecorationsForMatch = (
       }`
     : DecorationClassMap[DECORATION_MATCH];
 
+
+  const matchColour = match.markAsCorrect ? "3ff200" :  match.category.colour;
```
jonathonherbert, line 111: Nice-to-have: it'd be nice to apply this colour this with a class applied to the decoration, rather than hardcoding inline styles, as it would allow consumers to override the colour with CSS.

SHession: Yeah I totally agree

SHession: After discussing this further, @jonathonherbert and I feel this is acceptable as a temporary measure. The rationale for this is that categories owning custom colours is going to be replaced in favour of three standardised colours for Correct, Wrong and Possibly Wrong. When this happens we will adjust how the decoration styles are applied in favour of classes for each of these decorations. This will remove the inline styles, and at this point we can consider how a client may be able to customise these colours.

PR #86, opened 2020-08-06
Hide Advanced pane

tjsilver:
## What does this change?

The 'Advanced' pane is not required at present, and with the coming categories/tags refactor, it doesn't present useful things to the user.

I've commented out the code relating to the 'Advanced' pane and any CSS that was only used by that code. I chose to comment out the code (a) in case it might be used in the future, and (b) because some of the code wrapped already-commented-out code, but can delete instead if preferred.

## How to test

Run the branch locally and see that the 'Advanced' pane is no longer present.

### Before

image

image

### After

image

### DIFF
--- a/src/css/Sidebar.scss
+++ b/src/css/Sidebar.scss
@@ -41,29 +41,6 @@
   font-weight: bold;
 }
 
-// .Sidebar__header-toggle {
-//   cursor: pointer;
-// }
-
-// .Sidebar__toggle,
-// .Sidebar__toggle-label {
-//   align-self: baseline;
-//   font-size: $font-size-small;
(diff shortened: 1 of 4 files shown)


src/css/Sidebar.scss
```diff
@@ -41,33 +41,28 @@
...
 }
 
-.Sidebar__header-toggle {
-  cursor: pointer;
-}
-
-.Sidebar__header-option {
-  font-weight: regular;
-  font-size: $font-size-small;
-}
```
tjsilver, line 51: I deleted this completely as it wasn't being used anywhere.

PR #87, opened 2020-08-07
Add command to ignore matches

jonathonherbert:
## What does this change?

Adds a prosemirror command to ignore a match. Ignored matches are removed from the plugin state, along with their decorations.

PR for the relevant UI to follow.

## How to test

There's no UI path to exercise this code yet, but the tests should give some confidence that it works as expected.

Issuing the `ignoreMatches` command removes matches and match decorations from the plugin state.

### DIFF
--- a/src/ts/commands.ts
+++ b/src/ts/commands.ts
@@ -8,7 +8,8 @@ import {
   requestMatchesSuccess,
   requestError,
   requestMatchesForDirtyRanges,
-  requestMatchesComplete
+  requestMatchesComplete,
+  removeMatch
 } from "./state/actions";
 import {
   selectMatchByMatchId,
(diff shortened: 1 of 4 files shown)


src/ts/commands.ts
```diff
@@ -265,6 +266,23 @@ export const applyAutoFixableSuggestionsCommand = <TMatch extends IMatch>(
...
+  getState: GetState<TMatch>
+) => (
+  state: EditorState,
+  dispatch?: (tr: Transaction<any>) => void
+): boolean => {
+  const match = selectMatchByMatchId(getState(state), id);
+  if (match && dispatch) {
+    dispatch(state.tr.setMeta(PROSEMIRROR_TYPERIGHTER_ACTION, removeMatch(id)));
+  }
+  return !!match;
```
aug24, line 283: Why bother returning?  If it's idempotent, we don't care.

jonathonherbert: It'll tell the consumer whether it's found the match or not, as the function comment says ... having said that, nothing uses that API downstream, so arguably redundant.

jonathonherbert: For slightly more context – prosemirror commands have an API that makes dispatch optional, so we can run commands to see if something is possible without doing it – this would enable that behaviour to fit that convention. See 'Commands' in https://prosemirror.net/docs/guide/ for a description, this behaviour is what turns the menu icons in Composer on and off, for example.

PR #88, opened 2020-08-07
Add 'ignore match' functionality to UI

jonathonherbert:
NB: please review and merge #87 first! it splits the state legwork into a separate PR to make this more readable.

## What does this change?

Building on #87, this PR adds the 'ignore match' command to the UI.

tooltip-ignore

I've used material-ui to assist with buttons and tooltips here – the icons are already used in our other tooling, and I didn't want to reinvent the tooltip wheel when an implementation was already available in the package!

However, there are some problems with the preact implementation that prevent us using the icon components directly for now. I've used a copy-paste here for expediency, with a comment to let others know that this isn't desirable in the long term.

## How to test

In the plugin UI, clicking 'ignore this match' should remove the match from the document and the results pane.

As above.


build/rollup.config.dev.js
```diff
@@ -15,6 +18,15 @@ export default [
     },
     plugins: [
       ...defaultPlugins,
+      alias({
+        entries: {
+          react: "preact/compat",
```
jonathonherbert, line 23: We must alias `preact` to ensure material-ui imports `preact`, not `react`, when calling React-related functions. `preact` has a compatibility layer with an identical interface.

package.json
```diff
@@ -18,7 +18,7 @@
     "pages": "npm run build && http-server ./pages/dist",
     "prepublishOnly": "npm run build && tsc -p tsconfig.declaration.json",
     "build": "rollup -c build/rollup.config.build.js",
-    "watch": "rollup -w -c build/rollup.config.dev.js",
+    "watch": "rollup -w -m -c build/rollup.config.dev.js",
```
jonathonherbert, line 21: Add sourcemaps to aid debugging.

src/ts/components/Match.tsx
```diff
@@ -36,6 +41,20 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
       matchContext,
       markAsCorrect
     };
+
+    const suggestionsToRender = replacement ? [replacement] : suggestions || [];
```
jonathonherbert, line 45: A simple move to colocate var declarations.

src/ts/components/Match.tsx
```diff
@@ -48,19 +67,33 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
...
+                  href={this.getFeedbackLink(
+                    this.props.feedbackHref!,
+                    feedbackInfo
+                  )}
+                >
+                  Issue with this result? Tell us!
+                </a>
+              </div>
+            )}
+            <div className="MatchWidget__ignore-match">
```
jonathonherbert, line 84: The _actual_ tooltip and button markup!

aug24: Surely this shouldn't be displayed unless onIgnoreMatch is defined?

jonathonherbert: You're right! I've just made that change.

PR #90, opened 2020-08-07
Bundle material-ui with the dist build, ignoring other modules

jonathonherbert:
## What does this change?

At the moment, `material-ui` is not bundled with our dist build. This means that our 'preact' aliasing doesn't work, which breaks the build.

This PR adds material UI directly to the bundle. It comes at a significant bundle cost – 67kb to ~458kb – which we should work to mitigate as a priority. It may be that material-ui is not a suitable candidate for inclusion.

## How to test

Consume this plugin via e.g. `npm link` in a consumer project. It should work as expected.

### DIFF
--- a/build/rollup.config.build.js
+++ b/build/rollup.config.build.js
@@ -1,15 +1,39 @@
+import { nodeResolve } from "@rollup/plugin-node-resolve";
 import babel from "@rollup/plugin-babel";
-import replace from '@rollup/plugin-replace';
+import replace from "@rollup/plugin-replace";
 
 import { defaultPlugins } from "./rollup.common.js";
 
 const plugins = [
   ...defaultPlugins,
   babel(),
+  nodeResolve({ browser: true }),
   replace({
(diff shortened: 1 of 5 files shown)


build/rollup.config.build.js
```diff
@@ -8,24 +8,48 @@ const plugins = [
...
 export default [
   {
     input: "src/ts/index.ts",
     output: {
       file: "dist/index.js",
       format: "cjs"
     },
-    plugins
+    plugins,
+    external
```
jonathonherbert, line 44: By adding the rollup `commonjs` plugin to our bundle, the plugins listed in `external` are now eligible for bundling. (Previously, they were ignored). We only want to bundle material-ui.

We could look to automate this with something like [`rollup-plugin-peer-deps-external`](https://www.npmjs.com/package/rollup-plugin-peer-deps-external) in future, although arguably it's not a bad thing to be explicit in a library.

PR #91, opened 2020-08-07
Add an 'isActive' config param to the plugin, and remove decorations when the plugin is not active

jonathonherbert:
## What does this change?

Adds an 'isActive' config param to the plugin state. This removes decorations from the document when it's not `true`.

It anticipates a UI change to allow the active state to be toggled, which will change this param and respond to it by hiding the sidebar UI.

## How to test

There's no UI to test for this yet, but there are tests in the `createTyperighterPlugin` spec that test to check that decorations are applied to the view, or not.

We've an API to set the plugin to an active/inactive state.

## Dev notes

We have a few different configuration params, and PR adds another one, so I've grouped them into a `config` object and added a single, typesafe setter to reduce code duplication.

### DIFF
--- a/src/ts/commands.ts
+++ b/src/ts/commands.ts
@@ -3,8 +3,7 @@ import {
   newHoverIdReceived,
   requestMatchesForDocument,
   selectMatch,
-  setDebugState,
-  setRequestMatchesOnDocModified,
+  setConfigValue,
   requestMatchesSuccess,
   requestError,
   requestMatchesForDirtyRanges,
@@ -18,7 +17,8 @@ import {
 import {
   PROSEMIRROR_TYPERIGHTER_ACTION,
   IPluginState,
-  IStateHoverInfo
+  IStateHoverInfo,
+  IPluginConfig
 } from "./state/reducer";
 import {
   IMatcherResponse,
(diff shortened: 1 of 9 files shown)


src/ts/commands.ts
```diff
@@ -131,32 +131,20 @@ export const selectMatchCommand = <TMatch extends IMatch>(
...
-
-/**
- * When enabled, the plugin will queue match requests as soon as the document is modified.
- */
-export const setRequestOnDocModifiedState = (
-  requestMatchesOnDocModified: boolean
+export const setConfigValueCommand = <
+  ConfigKey extends keyof IPluginConfig,
+  ConfigValue extends IPluginConfig[ConfigKey]
+>(
```
jonathonherbert, line 139: I'm a big fan of this Typescript pattern – it ensures that
- the configuration keys are typesafe
- the configuration values are typesafe for that key

src/ts/createTyperighterPlugin.ts
```diff
@@ -106,7 +114,10 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
     },
     props: {
       decorations: state => {
-        return plugin.getState(state).decorations;
+        const pluginState = plugin.getState(state);
+        return pluginState.config.isActive
+          ? pluginState.decorations
+          : emptyDecorationSet;
```
jonathonherbert, line 120: This is where we decided to show/hide our document decorations – [`props.decorations`](https://prosemirror.net/docs/ref/#view.EditorProps.decorations) is a function which maps the stored plugin decorations to the decorations seen in the document.

src/ts/state/test/reducer.spec.ts
```diff
@@ -613,12 +614,34 @@ describe("Action handlers", () => {
...
+          setConfigValue("debug", true)
+        )
+      ).toEqual({ ...state, config: { ...state.config, debug: true } });
+    });
+    it("should not accept incorrect config keys", () => {
+      const { state } = createInitialData();
+      reducer(
+        new Transaction(createDoc),
+        state,
+        // @ts-expect-error
```
jonathonherbert, line 633: If you haven't come across `@ts-expect-error` before, you might have noticed – these tests don't assert anything! They're checking the typesafety of the `setConfigValue` function – [Typescript is expecting a type error there](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments), and fails if it doesn't see one. So the 'test' really happens at compile time.

src/ts/test/helpers/prosemirror.ts
```diff
@@ -15,3 +21,21 @@ const build = builders(schema, {
 
 export const createDoc = build.doc;
 export const p = build.p;
+
+/**
+ * We cannot compare document decorations to plugin decorations, as the
+ * .ProseMirror-widget class is applied – but we can compare their specs.
```
SHession, line 27: I still feel like I'm lacking knowledge in this area, and I would be interested to learn more about how decorations are applied in PM.

PR #92, opened 2020-08-07
Toggle on off

tjsilver:
## What does this change?

We want users to be able to turn Typerighter on and off in the UI so that they can work without distractions. This PR creates that functionality in the UI and links to state, following on from the work in this PR: https://github.com/guardian/prosemirror-typerighter/pull/91.

## How to test

Run locally and see the button to turn on Typerighter (it is defaulted to 'off' until linked with the store), and then click 'close' icon on the open sidebar to toggle.

The Typerighter UI is hidden when the 'close' icon is clicked and shown when 'open' is clicked.

### Before 
image

### After 
pm-on-off

#### Decorations hidden on close
decs-on-off

### DIFF
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -4,7 +4,8 @@ import Store, { STORE_EVENT_NEW_STATE } from "../state/store";
 import { IPluginState } from "../state/reducer";
 import { IMatch, ICategory } from "../interfaces/IMatch";
 import { selectHasError } from "../state/selectors";
-import { Button } from "@material-ui/core";
+// import IconButton from "@material-ui/core/IconButton";
+// import CloseIcon from "@material-ui/icons/Close";
 
 interface IProps {
   store: Store<IMatch>;
@@ -60,9 +61,16 @@ class Controls extends Component<IProps, IState> {
             >
               Check document
             </button>
-            <Button
+            <button
+              type="button"
+              className="Button Button__light"
               onClick={this.props.deactivate}
-            >Switch off</Button>
+            >Close</button>
+            {/* <IconButton
+              onClick={this.props.deactivate}
+            >
+              <CloseIcon fontSize="large" />
+            </IconButton> */}
           </div>
         </div>
       
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -63,6 +63,7 @@ class Controls extends Component<IProps, IState> {
             </button>
           
             <IconButton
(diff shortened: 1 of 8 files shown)


src/ts/components/Controls.tsx
```diff
@@ -58,6 +61,13 @@ class Controls extends Component<IProps, IState> {
             >
               Check document
             </button>
+          
+            <IconButton
+              aria-label="close Typerighter"
+              onClick={this.props.deactivate}
```
SHession, line 67: I think this icon would look a little better if it was set to small using the size prop: https://material-ui.com/api/icon-button/

src/ts/components/Sidebar.tsx
```diff
@@ -0,0 +1,65 @@
...
+                contactHref={contactHref}
+            />
+    </Fragment>
+        
+    ) : (
+        <button
+        type="button"
+        className="Button"
+        onClick={() => setIsActive(true)}
+    >Open Typerighter</button>
```
SHession, line 60: I think it would be nice if we make the switched off typerighter even less intrusive. Possibly similar to the Figma desing: https://www.figma.com/proto/mvbwYPVo3JCgQaP5APV0OK/Typerighter?node-id=42%3A1&scaling=min-zoom

src/ts/components/Controls.tsx
tjsilver: > The code changes look good, just a couple of comments on the UI. I would also favour Typerighter being active by default but I would be interested to hear your thoughts on that.

Yep, this is wired to start as 'off' for testing - it will be eventually set according to the isActive param in state when https://github.com/guardian/prosemirror-typerighter/pull/91/ is merged.

src/ts/components/Sidebar.tsx
tjsilver: 👍 working on that!

tjsilver:
> Tested this locally, looks great!
> 
> A couple of non-blocking comments:
> 
>     1. The open button might seem disabled due to it being grey.
> 
>     2. I wonder if it would be nice for the button text to simply read "Typerighter" instead of "Open Typerighter"



> Tested this locally, looks great!
> 
> A couple of non-blocking comments:
> 
>     1. The open button might seem disabled due to it being grey.
> 
>     2. I wonder if it would be nice for the button text to simply read "Typerighter" instead of "Open Typerighter"

Thanks Sam. Re (1), the button has the same styling as the 'check document' button, which in Composer is white with a grey hover state.

PR #94, opened 2020-08-10
Add an 'ignore matches' predicate

jonathonherbert:
## What does this change?

Adds an 'ignore matches' predicate that the plugin consumer can supply. This predicate receives every match supplied to the plugin on init or as a result of a successful response. If it returns false, the match is not added to the plugin state.

An example use case at the Guardian: in our text editor, our users have the ability to mark text ranges as 'correct'. We don't want to see matches for these ranges.

## How to test

The added tests should give us an idea of how this works.

You could also spin this up in a test environment and add a predicate to check it behaves as expected.

Matches that ... match the predicate are not included in the plugin state.

### DIFF
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -59,7 +59,7 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
     expandRanges = expandRangesToParentBlockNode,
     matches = [],
     isActive = true,
-    ignoreMatch = () => true
+    ignoreMatch = () => false
   } = options;
   // A handy alias to reduce repetition
   type TPluginState = IPluginState<TMatch>;
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -59,7 +59,7 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
     expandRanges = expandRangesToParentBlockNode,
     matches = [],
     isActive = true,
-    ignoreMatch
+    ignoreMatch = () => true
   } = options;
   // A handy alias to reduce repetition
   type TPluginState = IPluginState<TMatch>;
@@ -67,7 +67,7 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
   // Set up our store, which we'll use to notify consumer code of state updates.
   const store = new Store();
   const emptyDecorationSet = new DecorationSet();
-  const reducer = createReducer(expandRanges);
+  const reducer = createReducer(expandRanges, ignoreMatch);
 
   const plugin: Plugin = new Plugin({
     key: new PluginKey("prosemirror-typerighter"),
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -39,7 +39,7 @@ export interface IPluginOptions<TMatch extends IMatch = IMatch> {
   isActive?: boolean;
 
   /**
-   * Ignore matches when this predicate returns false.
+   * Ignore matches when this predicate returns true.
    */
   ignoreMatch?: IIgnoreMatch;
 }
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -1,5 +1,5 @@
 import { applyNewDirtiedRanges } from "./state/actions";
-import { IPluginState, PROSEMIRROR_TYPERIGHTER_ACTION } from "./state/reducer";
+import { IPluginState, PROSEMIRROR_TYPERIGHTER_ACTION, IIgnoreMatch } from "./state/reducer";
 import { createInitialState, createReducer } from "./state/reducer";
 import { selectNewBlockInFlight } from "./state/selectors";
 import { DECORATION_ATTRIBUTE_ID } from "./utils/decoration";
@@ -37,6 +37,11 @@ export interface IPluginOptions<TMatch extends IMatch = IMatch> {
    * Is the plugin active as it starts?
    */
   isActive?: boolean;
+
+  /**
+   * Ignore matches when this predicate returns false.
+   */
+  ignoreMatch?: IIgnoreMatch;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


SHession:
Would it be possible to get this rebased for a more accurate DIFF?

jonathonherbert:
@SHession great point, I've just done that 👍

src/ts/state/reducer.ts
```diff
@@ -146,25 +154,29 @@ export const PROSEMIRROR_TYPERIGHTER_ACTION = "PROSEMIRROR_TYPERIGHTER_ACTION";
...
+    dirtiedRanges: [],
+    currentMatches: [],
+    selectedMatch: undefined,
+    hoverId: undefined,
+    hoverInfo: undefined,
+    requestsInFlight: {},
+    requestPending: false,
+    requestErrors: []
+  };
+  return addMatchesToState(initialState, doc, matches, ignoreMatch);
```
jonathonherbert, line 176: The consumer can add matches when the plugin initialises – we need to run those matches through the `ignoreMatch` predicate, so a new helper function, `addMatchesToState`, takes care of that task.

src/ts/state/test/reducer.spec.ts
```diff
@@ -22,37 +22,63 @@ import {
...
+import { getBlocksFromDocument } from "../../utils/prosemirror";
 
-const reducer = createReducer(expandRangesToParentBlockNode);
+const reducer = createReducer(expandRangesToParentBlockNode, () => true);
+
+/**
+ * Create a plugin state, creating the given matches and
+ * their decorations from the given spec.
+ */
+const createStateWithMatches = (
```
jonathonherbert, line 46: There's a bit of scaffolding necessary to test the ignoreMatches predicate, which I've abstracted here. This is likely to be useful for other tests, but I haven't included that change here to keep the diff manageable.

src/ts/test/helpers/fixtures.ts
```diff
@@ -50,42 +50,69 @@ export const createBlock = (
...
-  blocks: [
-    {
-      id: createBlockId(0, from, to),
-      from,
-      to,
-      text: "block text"
-    }
-  ],
-  matches: [
+  specs: ICreateMatcherResponseSpec[],
```
jonathonherbert, line 64: I had to add a bit of detail to the fixture here to make it play nicely with the plugin state – we're testing more of it now, so we need a more accurate representation of what we'd get from a response.

src/ts/utils/decoration.ts
```diff
@@ -43,10 +43,7 @@ export const createDebugDecorationFromRange = (range: IRange, dirty = true) => {
 export const removeDecorationsFromRanges = (
   decorationSet: DecorationSet,
   ranges: IRange[],
-  types = [
-    DECORATION_MATCH,
-    DECORATION_MATCH_HEIGHT_MARKER
-  ]
+  types = [DECORATION_MATCH, DECORATION_MATCH_HEIGHT_MARKER]
```
jonathonherbert, line 46: Prettier. Very tempted add a lint step to our git hooks.

src/ts/createTyperighterPlugin.ts
```diff
@@ -53,20 +58,21 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
   const {
     expandRanges = expandRangesToParentBlockNode,
     matches = [],
-    isActive
+    isActive,
+    ignoreMatch = () => true
```
aug24, line 62: Err.  Shouldn't this be 'false'?

src/ts/state/helpers.ts
```diff
@@ -0,0 +1,22 @@
...
+import { createDecorationsForMatch } from "../utils/decoration";
+import { DecorationSet } from "prosemirror-view";
+
+export const addMatchesToState = <TMatch extends IMatch>(
+  state: IPluginState<TMatch>,
+  doc: any,
+  matches: TMatch[],
+  ignoreMatch: IIgnoreMatch
+) => {
+  const matchesToApply = matches.filter(ignoreMatch);
```
aug24, line 12: And surely this should be filter(!ignoreMatch) ?

src/ts/createTyperighterPlugin.ts
jonathonherbert: Yes. Yes it should.

src/ts/state/helpers.ts
jonathonherbert: As above 😅

src/ts/createTyperighterPlugin.ts
```diff
@@ -54,20 +59,21 @@ const createTyperighterPlugin = <TMatch extends IMatch>(
     expandRanges = expandRangesToParentBlockNode,
     matches = [],
     isActive = true,
+    ignoreMatch = () => false
```
aug24, line 62: `() => false` appears in four places.  Can it be DRYed out and given a name?

jonathonherbert: Great point, let's provide a default implementation.

src/ts/createTyperighterPlugin.ts
```diff
@@ -37,6 +37,11 @@ export interface IPluginOptions<TMatch extends IMatch = IMatch> {
    * Is the plugin active as it starts?
    */
   isActive?: boolean;
+
+  /**
+   * Ignore matches when this predicate returns false.
```
aug24, line 42: IS THIS COMMENT CORRECT?

src/ts/createTyperighterPlugin.ts
aug24: Did you?

src/ts/state/reducer.ts
```diff
@@ -92,6 +93,14 @@ export interface IBlockInFlight {
   block: IBlock;
 }
 
+/**
+ * A consumer-supplied predicate that allows consumers to ignore matches.
+ * Handy when, for example, consumers know that parts of the document are
+ * exempt from checks.
+ */
+export type IIgnoreMatch = (match: IMatch) => boolean;
+export const defaultIgnoreMatch: IIgnoreMatch = () => false;
```
aug24, line 102: You're going to hate me, but I don't like the name.  Why not 'includeAll' or something with business meaning?

jonathonherbert: > Did you?
I'm still blaming the heat 😅 

> You're going to hate me, but I don't like the name. Why not 'includeAll' or something with business meaning?
Naw, I completely agree. Like you say, business meaning > what the thing _is_. Just changed 👍

src/ts/createTyperighterPlugin.ts
jonathonherbert: ... now it is 😅

PR #95, opened 2020-08-12
Derive colours for matches

jonathonherbert:
_co-authored-by: @tjsilver_

## What does this change?

At the moment, we assign colours to rule matches in the UI via their categories.

This PR implements a new approach to colours for matches:

- Green: the tool knows this match is correct.
- Orange: the tool is unsure about this match, and would like the user to check it.
- Red: the tool is sure this match isn't correct, and has a replacement.

We can derive these colours from the match state. If a rule doesn't have a replacement, it's Orange; if a rule does have a replacement that's different from the matched text, it's Red; if a rule has a replacement that matches the matched text, it's green.

Before:

Screenshot 2020-08-12 at 12 17 38

After:

Screenshot 2020-08-12 at 12 16 09

## How to test

Run this plugin locally against text that has all three sorts of matches (for example, this [long read](https://www.theguardian.com/news/2019/aug/02/athleisure-barre-kale-tyranny-ideal-woman-labour).) You should see that match colours in the document, sidebar and match tooltips correspond to the above rules.

The above rules are implemented.

### DIFF
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -2,7 +2,7 @@ import { applyNewDirtiedRanges } from "./state/actions";
 import { IPluginState, PROSEMIRROR_TYPERIGHTER_ACTION } from "./state/reducer";
 import { createInitialState, createReducer } from "./state/reducer";
 import { selectNewBlockInFlight } from "./state/selectors";
-import { DECORATION_ATTRIBUTE_ID } from "./utils/decoration";
+import { DECORATION_ATTRIBUTE_ID, IMatchColours, defaultMatchColours } from "./utils/decoration";
 import { EditorView, DecorationSet } from "prosemirror-view";
 import { Plugin, Transaction, EditorState, PluginKey } from "prosemirror-state";
 import { expandRangesToParentBlockNode } from "./utils/range";
@@ -37,6 +37,11 @@ export interface IPluginOptions<TMatch extends IMatch = IMatch> {
    * Is the plugin active as it starts?
    */
   isActive?: boolean;
+
+  /**
+   * The colours to use for document matches.
+   */
+  matchColours?: IMatchColours;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/ts/utils/decoration.ts
```diff
@@ -145,8 +151,23 @@ export const createDecorationsForMatch = (
   return decorations;
 };
 
-export const createDecorationsForMatches = (matches: IMatch[]) =>
-  flatten(matches.map(_ => createDecorationsForMatch(_)));
+export const getColourForMatch = (
+  match: IMatch,
+  matchColours: IMatchColours
+) => {
```
jonathonherbert, line 157: This expresses the logic in the PR description.

src/ts/createTyperighterPlugin.ts
```diff
@@ -37,6 +37,11 @@ export interface IPluginOptions<TMatch extends IMatch = IMatch> {
    * Is the plugin active as it starts?
    */
   isActive?: boolean;
+
+  /**
+   * The colours to use for document matches.
+   */
+  matchColours?: IMatchColours;
```
jonathonherbert, line 44: We add this as a configuration value, so plugin consumers can provide their own colours. This will be important for us, as well as 3rd party users – we don't want to have to rerelease when our colour scheme changes!

src/ts/utils/decoration.ts
```diff
@@ -98,21 +108,17 @@ const createHeightMarkerElement = (id: string) => {
...
     : DecorationClassMap[DECORATION_MATCH];
 
-
-  const matchColour = match.markAsCorrect ? "3ff200" :  match.category.colour;
+  const matchColour = getColourForMatch(match, matchColours);
   const opacity = isSelected ? "30" : "07";
-  const style = `background-color: #${
-    matchColour
-  }${opacity}; border-bottom: 2px solid #${matchColour}`;
+  const style = `background-color: ${matchColour}${opacity}; border-bottom: 2px solid ${matchColour}`;
```
SHession, line 121: Did we want to change this to apply specific classes instead of inline styles?

jonathonherbert: As you mention below, the colours can be fed in by the consumer in `createTyperighterPlugin`, so we've an API to modify them that doesn't require classes.

PR #96, opened 2020-08-12
Reduces the opacity of ambiguous decorations

SHession:
## What does this change?

This PR reduces the opacity of the ambiguous colour, this reduces the amount of visual noise that Typerighter creates for rules that may be inaccurate. This PR is based on the work done to [change the match colours](https://github.com/guardian/prosemirror-typerighter/pull/95).

## How to test

Open a document, run the document through Typerighter, observe the amount of visual noise generated by ambiguous rules. 

Users' attention is on matches which are most significant and accurate.

## Images

Before:
Screenshot 2020-08-12 at 14 08 00

After:
Screenshot 2020-08-12 at 14 07 21

### DIFF
--- a/src/ts/utils/decoration.ts
+++ b/src/ts/utils/decoration.ts
@@ -116,9 +116,8 @@ export const createDecorationsForMatch = (
     ? `${DecorationClassMap[DECORATION_MATCH]} ${DecorationClassMap[DECORATION_MATCH_IS_SELECTED]}`
     : DecorationClassMap[DECORATION_MATCH];
 
-  const matchColour = getColourForMatch(match, matchColours);
-  const opacity = isSelected ? "30" : "07";
-  const style = `background-color: ${matchColour}${opacity}; border-bottom: 2px solid ${matchColour}`;
+  const {backgroundColour, borderColour} = getColourForMatch(match, matchColours, isSelected);
+  const style = `background-color: ${backgroundColour}; border-bottom: 2px solid ${borderColour}`;
 
   const decorations = [
     Decoration.inline(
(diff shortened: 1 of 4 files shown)


src/ts/utils/decoration.ts
```diff
@@ -151,17 +150,30 @@ export const createDecorationsForMatch = (
...
-    return matchColours.unambiguous;
+    return {
+      backgroundColour: `${matchColours.unambiguous}${backgroundOpacity}`, 
+      borderColour: `${matchColours.unambiguous}`
+    };
   }
-  return matchColours.ambiguous;
+  return {
+    backgroundColour: `${matchColours.ambiguous}${backgroundOpacity}`, 
+    borderColour: `${matchColours.ambiguous}4D`
```
tjsilver, line 175: `nit` Could this be pulled out into a borderOpacity const for clarity?

SHession:
> I think the styling changes make loads of sense. But I'd prefer to keep opinions about styling of this sort out of the plugin, if possible. For example, it'd be painful to issue another release just to tweak the opacity of the background.
> 
> With this in mind, could we add both background and border colour explicitly to the `IMatchColours` interface? In this way, we can set defaults as you have them above, but also allow users to explicitly set them when they create the plugin.
> 
> That way, in Composer, if we want to adjust, it's a single PR, rather than PR -> release -> consumer PR.

I agree with the first part. 

In regards to the the second part, are you suggesting we allow the consumer to specify a background and border colour for each state? Or are you suggesting we allow the consumer to specify the opacity variables, such as selected opacity or ambiguous opacity?

jonathonherbert:
> In regards to the the second part, are you suggesting we allow the consumer to specify a background and border colour for each state? Or are you suggesting we allow the consumer to specify the opacity variables, such as selected opacity or ambiguous opacity?

I think the former, because it's the simplest way to avoid having to make changes to plugin code in response to user requirements. It's slightly more verbose, but by supplying sensible defaults we can avoid needing to specify these values downstream until our opinions about styling change.

SHession:
> I think the former, because it's the simplest way to avoid having to make changes to plugin code in response to user requirements. It's slightly more verbose, but by supplying sensible defaults we can avoid needing to specify these values downstream until our opinions about styling change.
> 
> EDIT: to clarify, this isn't a very strongly held opinion – either would be fine 👍
> 
> (and I hadn't thought about adjusting e.g. selection opacity!)

Is the aim of this to include the opacity in the provided colours? In which case we would have to say don't provide the opacity for the background colours but do for the border colours, as the background colours have their opacity applied based on isSelected. Might it be best to simply provide an ambiguous opacity variable? The control over selection opacity might not be very useful.

PR #97, opened 2020-08-13
Add interfaces to exports

jonathonherbert:
## What does this change?

We've added a few predicates/callbacks recently (#94, #88) that require us to reason about plugin interfaces in consumer code. We'll need to export those interfaces so consuming code can use them when defining those predicates.

## How to test

This is a no-op.

We can access the given interfaces in consumer code.

### DIFF
--- a/src/ts/state/store.ts
+++ b/src/ts/state/store.ts
@@ -100,6 +100,6 @@ class Store<
   }
 }
 
-export type IStore = typeof Store;
+export type IStore<TMatch extends IMatch = IMatch> = Store<TMatch>;
 
 export default Store;
--- a/src/ts/state/store.ts
+++ b/src/ts/state/store.ts
@@ -100,4 +100,6 @@ class Store<
   }
 }
 
+export type IStore = typeof Store;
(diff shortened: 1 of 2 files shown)


src/ts/state/store.ts
```diff
@@ -100,4 +100,6 @@ class Store<
   }
 }
 
+export type IStore<TMatch extends IMatch = IMatch> = Store<TMatch>;
```
SHession, line 103: Should this be done in a new file in the interface folder, or would that be overkill?

jonathonherbert: It's a great point. We're quite inconsistent in this project – sometimes interfaces are defined inline, sometimes they have their own files. One reason is that if an interface isn't clearly 'owned' by a piece of code (e.g. match and block interfaces), it's difficult to know which code to colocate it with. For other interfaces (e.g. `IStore` above) it's clearer where to put things.

It'd be nice to establish a convention, document it, refactor where necessary and stick with it. That change probably wants its own PR.

PR #105, opened 2020-08-14
Trigger search on toggle

tjsilver:
## What does this change?
At the moment, users must click to activate Typerighter and click again to check a document. Re-activating the plugin without rechecking might be confusing to users. This PR addresses this by keeping the 'Check document' button across open/closed states of typerighter. As soon as 'Check document' is clicked in the closed state, a check is triggered and Typerighter is activated.

## How to test

Run branch locally and observe that Typerighter carries out a document check when 'Check document' is clicked in the closed state.

Users will find the behaviour intuitive.

### Before
toggle-on-off2

### After
toggle-on-off

### DIFF
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -17,12 +17,11 @@ interface IProps {
   addCategory: (id: string) => void;
   removeCategory: (id: string) => void;
   feedbackHref?: string;
-  deactivate: () => void;
+  onToggleActiveState: () => void;
 }
 
 interface IState {
   pluginState: IPluginState<IMatch> | undefined;
-  isOpen: boolean;
   allCategories: ICategory[];
   currentCategories: ICategory[];
   isLoadingCategories: boolean;
@@ -33,7 +32,6 @@ interface IState {
  */
 class Controls extends Component<IProps, IState> {
   public state = {
-    isOpen: false,
     allCategories: [],
     currentCategories: [],
     isLoadingCategories: false,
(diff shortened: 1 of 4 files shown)


src/ts/components/Controls.tsx
```diff
@@ -33,45 +32,58 @@ interface IState {
...
 class Controls extends Component<IProps, IState> {
   public state = {
-    isOpen: false,
     allCategories: [],
     currentCategories: [],
     isLoadingCategories: false,
-    pluginState: undefined,
+    pluginState: undefined
   } as IState;
   public componentWillMount() {
```
SHession, line 40: We should aim to remove componentWillMount whenever possible as it's now deprecated. https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

src/ts/components/Controls.tsx
```diff
@@ -33,45 +32,58 @@ interface IState {
...
     this.props.store.on(STORE_EVENT_NEW_STATE, this.handleNotify);
+    this.setState({ pluginState: this.props.store.getState() });
     this.initCategories();
   }
 
   public render() {
+    const checkInProgress: boolean | undefined =
+      this.state.pluginState &&
```
SHession, line 48: I think accessing this plugin state could be better achieved through the usage of selectors.

jonathonherbert:
Great point re: lifecycle and selectors @SHession – a hooks refactor would be nice at some point, and I'd love to get `react-redux` into this project to make mapping from plugin state -> component props more transparent (and performant!), at which point enforcing selectors in `mapStateToProps` becomes easier to manage.

PR #106, opened 2020-08-14
Offset API for 'scroll to rule' behaviour 

jonathonherbert:
## What does this change?

This PR adds an offset API for the 'scroll to rule' sidebar behaviour. The example implementation centres the rules in the editor.

scroll-to-match

## Implementation notes

Because we don't know much about our consumer's editor, we defer to them with a `getScrollOffset` callback.

It's a callback, rather than a static value, because the editor size may change during the page lifecycle (for example, if the user alters the window size), and it's likely we'd need to adjust accordingly to e.g. centre matches correctly.

## How to test

In the example implementation, click a rule in the sidebar in a long-ish document. The editor should scroll to the match in the document, centering it on the page.

Plugin consumers have more control about how the plugin scrolls matches into view.


src/ts/utils/decoration.ts
```diff
@@ -201,3 +201,7 @@ export const findSingleDecoration = (
   }
   return decorations[0];
 };
+
+export const maybeGetDecorationElement = (matchId: string): HTMLElement | null => document.querySelector(
```
jonathonherbert, line 205: I figure it's better to colocate this with the decoration code.

pages/index.ts
```diff
@@ -50,6 +50,10 @@ if (editorElement && sidebarNode) {
     })
   });
 
+  // When the user scrolls to matches, place the match in the middle of the editor.
+  const menuHeight = 47;
```
SHession, line 54: Where does this come from /Is there a nicer way to get this value?

SHession: I withdraw the question, as it's only related to the demo implementation.

jonathonherbert: Haha, yeah, that was my thought 😁

PR #107, opened 2020-08-14
Mark as correct

SHession:
## What does this change?

This PR adjusts the styling for the "Mark as correct" option (formally ignore match). It moves the functionality from an icon in the bottom left corner, to where the suggestions are normally displayed. 

## How to test

Check a document for matches, hover over a match, observe the "Mark as correct" option and behaviour.

Users can easily and intuitively dismiss a match by confirming it's correct.

## Images
After:

TyperighterMark

Before: 
image

### DIFF
--- a/src/ts/createView.tsx
+++ b/src/ts/createView.tsx
@@ -17,7 +17,7 @@ interface IViewOptions {
   contactHref?: string;
   feedbackHref?: string;
   logger?: ILogger;
-  onIgnoreMatch?: (match: IMatch) => void;
+  onMarkCorrect?: (match: IMatch) => void;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/ts/createView.tsx
```diff
@@ -59,11 +59,11 @@ const createView = ({
...
         commands.stopHover();
       }}
-      onIgnoreMatch={
-        onIgnoreMatch &&
+      onMarkCorrect={
+        onMarkCorrect &&
         (match => {
           commands.ignoreMatch(match.matchId);
-          onIgnoreMatch(match);
+          onMarkCorrect(match);
```
SHession, line 66: There is now a difference between the callback name and the command. I think this is acceptable but if anyone feels differently, please let me know.

PR #112, opened 2020-08-18
Use popper.js to position tooltips properly

jonathonherbert:
## What does this change?

Previously, some ad-hoc position code was being used to position tooltips. This is a hard, edge-casey, solved problem. This PR uses [popper.js](https://popper.js.org/) to solve it instead.

## How to test

Tooltips should appear underneath match decorations as usual, but now, they're liberated from the bounds of their editor element.

Screenshot 2020-08-18 at 18 05 30

Tooltips are no longer hidden by other DOM elements or constrained by the editor element.

### DIFF
--- a/build/rollup.config.build.js
+++ b/build/rollup.config.build.js
@@ -29,7 +29,9 @@ const externalModules = [
   "prop-types",
   "react-is",
   "hoist-non-react-statics",
-  "diff"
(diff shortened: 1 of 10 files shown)


SHession (approved):
Tested locally and reviewed the code. Works nicely!

PR #113, opened 2020-08-18
Remove redundant tooltip code

jonathonherbert:
## What does this change?

This PR cleans up after #112. Because we're delegating to an external lib to manage tooltips, all of the tooltip state we we storing, and the code that generates it, can go.

In its place, we have two sets of commands relating to two states:

- `hover` indicates the user is hovering over a match (which we use to conditionally show tooltips)
- `highlight` indicates we should show a focus state for the given match, without the tooltip.

## How to test

This is a no-op – everything should work as expected (see especially tooltip and sidebar highlight behaviour).


src/ts/commands.ts
```diff
@@ -75,35 +75,66 @@ export const requestMatchesForDirtyRangesCommand = (
...
+    );
+  }
+  return true;
+};
+
+
+/**
+ * Indicate the user is hovering over a match.
+ */
+export const startHighlightCommand = (matchId: string): Command => (
```
jennygrahamjones, line 114: Is this command is, like `startHoverCommand`, intended to indicate a user is hovering over a match? The comment says it is, but I am not clear what is the difference between the two. I think I'm lacking context for the terminology here, perhaps.

jonathonherbert: @jennygrahamjones Brilliant spot, I'd seen this and forgotten to amend – the two states are different. Does the updated comment make more sense?

jennygrahamjones (approved):

PR #114, opened 2020-08-19
Switch to react

jonathonherbert:
## What does this change?

In the Guardian tools that consume this product, React is ubiquitous.

In addition to this, using Preact forces us to bundle third-party UI code in the distribution build, or for consumer code to handle Preact aliasing – even if they don't use Preact themselves. There are also problems importing some modules with aliasing, e.g. the material-ui icons, which we've had to manually reconstitute ourselves.

So the cost of including Preact I think outweights the benefit of the ~33kb we save on bundle size, which in real terms has not yet affected our users.

This PR removes Preact in favour of React for the reasons above.

## How to test

This is a no-op – the tool should behave exactly as before.

There's one exception – in using material-ui icons, the 'mark as correct' icon has changed. Let me know if that's not right! (cc. @SHession)

Screenshot 2020-08-19 at 08 22 32

## Dev notes

There are many changes here, but I've tried to keep the commit history clean so you can see the steps I've taken – it's probably easiest to verify the sorts of changes that have occurred on a commit basis.

The majority of the changes to components are
- adjusting imports
- renaming the `class` prop to `className`
- removing props from the `render` function to `this.props` (in Preact they're passed to render directly)


tjsilver (approved):
This is great! Tested locally and works well.

PR #115, opened 2020-08-19
Add markdown parsing to the match descriptions

jonathonherbert:
## What does this change?

Adds markdown to the match descriptions.

Before:

Screenshot 2020-08-19 at 10 31 16

After:
Screenshot 2020-08-19 at 09 04 54

## How to test

Trigger a match with markdown syntax, like the example. It should be rendered correctly.

## Dev notes

Because we don't expect HTML markup, we can avoid using a sanitisation library by stripping HTML from the message data altogether – see `stripHtml`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "@popperjs/core": "^2.4.4",
     "diff": "^4.0.2",
     "lodash": "^4.17.11",
+    "marked": "^1.1.1",
     "preact": "^10.4.7",
     "prosemirror-tables": "^0.7.8",
     "prosemirror-utils": "^0.6.7",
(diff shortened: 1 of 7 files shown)


jennygrahamjones (approved):

package.json
```diff
@@ -86,6 +86,7 @@
     "prosemirror-tables": "^0.7.8",
     "prosemirror-utils": "^0.6.7",
     "react-popper": "^2.2.3",
+    "snarkdown": "^1.2.2",
```
jennygrahamjones, line 89: What a neat little parser!

jonathonherbert: Isn't it? I love the package description: `basically one regex and a huge if statement`

PR #117, opened 2020-08-19
Ts/error on auth fail

tjsilver:
Co-authored-by: Jonathon Herbert <jonathon.herbert@guardian.co.uk>

## What does this change?
Currently, when there is an authentication error, the user sees an error message stating 'error fetching matches...'. This PR is to ensure the Typerighter plugin gives a clear error on authentication failures, so the user is prompted to refresh the page. This PR also addresses a UI bug which meant that error messages persisted when the plugin was closed.

## How to test

Run the branch locally, delete any cookies on Composer that are allowing auth of Typerighter, and see an error message suggesting you refresh the page. 

Users should be able to differentiate between authentication errors and errors in fetching matches in Typerighter.

### Before - all errors
auth-error-before

### After - auth errors

auth-error

### After - match errors

auth-error-after2

### DIFF
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -6,7 +6,11 @@ import { Close } from "@material-ui/icons";
 import Store, { STORE_EVENT_NEW_STATE } from "../state/store";
 import { IPluginState } from "../state/reducer";
 import { IMatch, ICategory } from "../interfaces/IMatch";
-import { selectHasError, selectRequestsInProgress } from "../state/selectors";
+import {
+  selectHasError,
+  selectHasAuthError,
+  selectRequestsInProgress
+} from "../state/selectors";
 
 interface IProps {
   store: Store<IMatch>;
@@ -45,7 +49,6 @@ class Controls extends Component<IProps, IState> {
   }
 
   public render() {
-
     const handleCheckDocumentButtonClick = (): void => {
       if (!this.state.pluginState?.config.isActive) {
         this.props.onToggleActiveState();
@@ -65,7 +68,10 @@ class Controls extends Component<IProps, IState> {
               type="button"
               className="Button"
               onClick={handleCheckDocumentButtonClick}
-              disabled={this.state.pluginState && selectRequestsInProgress(this.state.pluginState)}
+              disabled={
+                this.state.pluginState &&
+                selectRequestsInProgress(this.state.pluginState)
+              }
             >
               Check document
             </button>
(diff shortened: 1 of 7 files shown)


src/ts/components/Controls.tsx
```diff
@@ -57,6 +62,43 @@ class Controls extends Component<IProps, IState> {
...
+
+      let errorMessage: string = "";
+      if (hasAuthError) {
+        errorMessage = "Authentication error - please refresh the page. ";
+      }
+      if (hasGeneralError && !hasAuthError) {
+        errorMessage =
+          "Error fetching matches. Please try checking the document again. ";
+      }
+      if (hasErrors) {
```
jonathonherbert, line 84: Could we return earlier here (e.g. L75), to save us trying to make up the error message when no errors exist?

src/ts/components/Controls.tsx
```diff
@@ -57,6 +62,43 @@ class Controls extends Component<IProps, IState> {
...
+
+      if (!pluginState) {
+        return;
+      }
+
+      const hasAuthError = selectHasAuthError(pluginState);
+      const hasGeneralError = selectHasGeneralError(pluginState);
+      const hasErrors: boolean = hasAuthError || hasGeneralError;
+
+      let errorMessage: string = "";
```
jonathonherbert, line 76: Because we know we've only got two kinds of error, could this read:
```
const errorMessage = hasAuthError ? "...auth error msg" : "normal error msg"
```
We'd need to refactor this anyway if we added additional error types.

(We'd need to return earlier if there weren't any errors, re: comment below, for this to work)

src/ts/state/selectors.ts
```diff
@@ -123,11 +110,19 @@ export const selectAllAutoFixableMatches = <TMatch extends IMatch>(
...
   );
 
-export const selectHasError = <TMatch extends IMatch>(
+export const selectHasGeneralError = <TMatch extends IMatch>(
   state: IPluginState<TMatch>
-): boolean =>
-  !!state.requestErrors && state.requestErrors.length > 0;
+): boolean => {
+  const generalErrors = state.requestErrors.filter(_ => _.type === "GENERAL_ERROR");
+  return !!state.requestErrors && generalErrors.length > 0};
```
jonathonherbert, line 117: Is the `!!state.requestErrors` check necessary? Looking at the `IState` type, it looks like this will always be true, as it's a non-optional array? (Same for `selectHasAuthError` below)

jonathonherbert (approved):
Tested locally and works great. A couple of refactor suggestions but nothing blocking 👍

PR #123, opened 2020-08-26
Refactor more class components

tjsilver:
## What does this change?
The Results and SidebarMatch components were React class components. In this PR they have been refactored into functional components. This allows us to use the`useContext` hook and has the additional benefit of removing any deprecated lifecycle methods such as `componentWillMount`.

## How to test
Run the branch locally and see that the sidebar controls and results work as expected.


jonathonherbert (approved):
Works as expected 👍

PR #125, opened 2020-08-27
Set up telemetry service

tjsilver:
## What does this change?
We will have an analytics client shortly, responsible for batching and submitting telemetry events. We'd like our Prosemirror plugin to send events when they occur, via that interface.

This PR implements an events service with prosemirror-typerighter that calls the appropriate code when events happen. The backend of the implementation logs to console for now.

## How to test
Run the branch locally and check the console for details of events triggered. You should see a separate log entry for:
- Checking document
- Using 'close' button
- Clicking on a sidebar match
- Accepting a suggestion
- Marking a match as correct.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -16,6 +16,8 @@ import createView from "../src/ts/createView";
 import { createBoundCommands } from "../src/ts/commands";
 import MatcherService from "../src/ts/services/MatcherService";
 import { TyperighterAdapter } from "../src/ts";
+import { ITyperighterTelemetryEvent } from "../src/ts/interfaces/ITelemetryData";
+import TelemetryService from "../src/ts/services/TelemetryService";
 
 const mySchema = new Schema({
   nodes: addListNodes(schema.spec.nodes as any, "paragraph block*", "block"),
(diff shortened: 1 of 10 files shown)


pages/index.ts
```diff
@@ -69,6 +71,11 @@ if (editorElement && sidebarNode) {
     commands,
     new TyperighterAdapter("https://api.typerighter.local.dev-gutools.co.uk")
   );
+
+  const stubTelemetrySender = (event: ITyperighterTelemetryEvent) =>
+  console.log(event);
+  const telemetryService = new TelemetryService(stubTelemetrySender);
```
tjsilver, line 77: This is here for testing.

src/ts/components/Controls.tsx
```diff
@@ -85,10 +88,18 @@ const Controls = ({
     const handleCheckDocumentButtonClick = (): void => {
       if (!pluginIsActive) {
         onToggleActiveState();
+        telemetryService?.typerighterIsOpened({documentUrl: document.URL})
```
jonathonherbert, line 91: This is the perfect use-case for optional chaining – love to see it.

jonathonherbert (approved):
Code looks great, works as expected – great to see those very official looking events in the console (for now!)

PR #126, opened 2020-09-07
Preserve marks when issuing replacements – step 1

jonathonherbert:
## What does this change?

At the moment, when suggestions are applied to a document, they replace the entire range they're applied to with a new text node. This means that marks that applied to the previous range – e.g. bold styling, or link information – is lost.

This PR copies marks across when they span the whole of the range altered by the selection, preserving them in some, but not all circumstances.

Before

preserve-marks-1-before

After

preserve-marks-1-before

This is better, but not quite ideal – marks within the selection that don't cover it will still be lost. For example, the applying the suggestion 'Example' to the text 'exa**mple**' will remove the bold stying, as it doesn't cover the whole range.

We can address this in a follow up PR by diffing the suggested text against what's being replacement, and only applying the above logic to the diff – in the above, only the replacement `e` -> `E` would be applied, preserving marks in the remainder of the range.

## How to test

Spin up an instance of the plugin locally. Does it behave as expected.

The automated tests should pass, and the editor should behave as expected when suggestions are accepted when marks cover the whole of the affected range.

## Notes

The change here is quite simple, but difficult to test without running the entire editor and querying the DOM rather than e.g. plugin or editor state. They've been needed for a while, so I've added them in this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -772,6 +772,24 @@
         }
       }
     },
+    "@babel/runtime-corejs3": {
+      "version": "7.11.2",
+      "resolved": "https://registry.npmjs.org/@babel/runtime-corejs3/-/runtime-corejs3-7.11.2.tgz",
+      "integrity": "sha512-qh5IR+8VgFz83VBa6OkaET6uN/mJOhHONuy3m1sgF0CV6mXdPSEBdA7e1eUbVvyNtANjMbg22JUv71BaDXLY6A==",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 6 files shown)


SHession (approved):
Reviewed the code and tested locally, looks great. The tests are a great addition, but might be nice to have more test cases with different marks or multiple marks.

jonathonherbert:
Thanks @SHession! Re: tests, the next PR should test and address many more edge cases 👍

PR #127, opened 2020-09-08
Preserve marks when issuing replacements – step 2

jonathonherbert:
**NB:** depends on #126. Please review that first!

## What does this change?

Builds on #126 to do a more complete job of preserving marks when replacements are issued. Previously, the entire range covered by the match was replaced by the suggestion.

In this PR, we create a diff between the current and replacement text, and only intervene in the document when text is changed. For additions, we copy the marks that cover the range to be replaced.

For example, when applying the suggestion 'Example' to the text '**ex**ample', only the replacement e -> E would be applied:

- the bold '**x**' is not touched, as it's not part of the diff
- the bold '**e**' is marked for replacement, and its mark is copied across to its replacement, 'E'.

There is also a heuristic to preserve styling when characters are added. For example, when changing **eggs** -> beggars, we'd expect the entire replacement to also be bold. In this implementation, the patches 'b' and 'ars' inherit marks – 'b' because the mark is inclusive, and 'ars' because we look back one character and inherit the marks from there.

Before:

preserve-marks-2-after

After:

preserve-marks-1-before

## How to test

Apply suggestions to a document. They should preserve the styling of their replaced ranges insofar as possible, bearing in mind the methods we're using above.

Users no longer feel that applying suggestions removes their styling from the replaced ranges.


src/ts/commands.ts
```diff
@@ -322,26 +321,35 @@ const maybeApplySuggestions = (
...
+    return true;
+  }
+
+  const tr = state.tr;
+  suggestionsToApply.forEach(({ from, to, text }) => {
+    if (!text) {
+      return;
+    }
+    const mappedFrom = tr.mapping.map(from);
+    const mappedTo = tr.mapping.map(to);
```
SHession, line 334: `nit` Should these consts live inside the `getReplacementFragmentsFromReplacement` function?

src/ts/utils/prosemirror.ts
```diff
@@ -75,3 +78,101 @@ export const getReplaceTransactions = (tr: Transaction) =>
...
+
+      const prevFragment = currentFragments[currentFragments.length - 1];
+      const isThisPatchNew =
+        !!prevFragment && (prevFragment.to || 0) < $from.pos;
+
+      let marks;
+      if (isThisPatchNew) {
+        // If this patch adds characters and the previous patch left
+        // a range intact, inherit the marks from the last character
+        // of that range.
```
SHession, line 150: I'm struggling to decipher this comment, could you try explaining it another way? What is meant by leaving a range intact?

src/ts/test/commands.spec.ts
```diff
@@ -34,5 +34,53 @@ describe("Commands", () => {
...
     });
+
+    it("should keep marks across the whole replaced text when suggestions are applied and additions are made to the beginning of the range", () => {
+      const match = createMatch(5, 9, [
+        { text: "beggar", type: "TEXT_SUGGESTION" }
+      ]);
+      const {
+        editorElement,
+        commands
+      } = createEditor("<p>Two <strong>eggs</strong></p>", [match]);
```
SHession, line 45: `nit` It would be nice to see a test when some text is bold and italicised, to prove it can handle multiple marks.

SHession (approved):
I've tested this locally and reviewed the code. It does a decent job of preserving marks! 
I've added a few minor comments for your consideration.

src/ts/utils/prosemirror.ts
jonathonherbert: Yes, this is an interesting one to explain – I've just updated that comment. Let's discuss further if that's not clear, it'd be nice to make this as comprehensible as possible!

src/ts/commands.ts
jonathonherbert: Great spot – I was unsure whether this was necessary when first implementing!

jonathonherbert:
@SHession I've refactored to make this code a bit more comprehensible and address the edge case you discovered 👍

src/ts/utils/prosemirror.ts
```diff
@@ -122,45 +143,53 @@ export const getReplacementFragmentsFromReplacement = (
...
-      const $from = tr.doc.resolve(from + currentPos);
-      const $to = tr.doc.resolve(Math.min($from.pos + patch.count, tr.doc.nodeSize - 2));
+
+      const newFrom = from + currentPos;
+      const newTo = newFrom + patch.count;
 
       // If this patch removes chars, create a fragment for
       // the range, and leave the cursor where it is.
       if (patch.removed) {
+        const fragment = {
```
SHession, line 153:
```suggestion
        const fragment : : ISuggestionPatchDelete = {
```
`nit` Is this a nicer way to create an instance, as it means you don't need the `as const`?

src/ts/utils/prosemirror.ts
```diff
@@ -122,45 +143,53 @@ export const getReplacementFragmentsFromReplacement = (
...
+      const isInsertion = !!prevFragment && (prevFragment.to || 0) < newFrom;
+
+      let getMarks;
+      if (isInsertion) {
+        // If this patch is an insertion, inherit the marks from the last character
+        // of that patch's range. This ensures that text that expands upon an existing
+        // range shares the existing range's style.
+        getMarks = (localTr: Transaction) => {
+          const $newFrom = localTr.doc.resolve(newFrom)
+          const $lastCharFrom = localTr.doc.resolve(newFrom - 1);
```
SHession, line 175: I was a bit confused at how this would work in all cases, until I realised the **b** in **eggs** -> **beggars** doesn't count as an `isInsertion`. Perhaps a small comment about this on line :166 would be helpful.

SHession (approved):
This looks like its working well. I can confirm "I’m a Celebrity … Get Me Out of Here!" now correctly has its marks preserved.

I have added a couple of small suggestion. Additionally, one other suggestion for a test that might be nice, is one with overlapping marks. However it's not a blocker.

src/ts/utils/prosemirror.ts
jonathonherbert: Yeah, `isInsertion` doesn't cover exactly what's happening here – I've updated the name of the var, and the comment, to hopefully do a better job! 👍

jonathonherbert:
@SHession great suggestions. I've added that test, and added a utility to make suggestion tests easier to write.

PR #128, opened 2020-09-09
Create telemetry client

jonathonherbert:
_Co-authored-with: @tjsilver_ 

## What does this change?

Adds a telemetry client to send telemetry events to our [remote telemetry service](https://github.com/guardian/editorial-tools-user-telemetry-service).

The telemetry client is responsible for aggregating events and sending them to a remote service on a throttled basis, retrying if requests fail.

The intent is to eventually house this client within the remote telemetry service and release it as a package, but to speed up the process of getting it into users' hands, we've added it here for now.

## How to test

In a local instance, perform actions that trigger telemetry events – like opening/closing Typerighter, checking the document, or accepting suggestions. On a throttle of 10s you should see events sent to the telemetry service.

Telemetry events begin to populate our telemetry service.

We limit the number of events we send to a total of 500 in a single payload. Remaining events are put back in the buffer and picked up on the next throttle pass.

We haven't yet implemented an exponential backoff for this service, which would be good to have as usage grows.


package.json
```diff
@@ -34,7 +35,7 @@
     "@rollup/plugin-replace": "^2.3.3",
     "@testing-library/dom": "^7.23.0",
     "@types/diff": "^4.0.2",
-    "@types/fetch-mock": "^6.0.4",
+    "@types/fetch-mock": "^7.3.3",
```
jonathonherbert, line 38: It's important these match!

src/ts/services/TelemetryService.ts
```diff
@@ -1,69 +1,55 @@
...
 
-  public sidebarMatchClicked(tags: ISidebarClickEvent["tags"]) {
-    this.sendTelemetryEvent({
-        type: TYPERIGHTER_TELEMETRY_TYPE.TYPERIGHTER_SIDEBAR_MATCH_CLICK,
-        value: 1,
-        eventTime: new Date().toISOString(),
-        tags
-    });
-  }
+        const response = await fetch(`${this.telemetryUrl}/event`, {
```
SHession, line 23: Would it be better not to assume a certain endpoint is on the url and trust the the parameter parsed in is correct?

src/ts/createView.tsx
```diff
@@ -28,7 +28,7 @@ interface IViewOptions {
   // to place the match in the middle of the screen, as the size of the
   // document might change during the lifecycle of the page.
   getScrollOffset?: () => number;
-  telemetryService?: TelemetryService;
+  telemetryService?: TyperighterTelemetryAdapter;
```
SHession, line 31: Should the name of this property instead be telemetry adapter?

src/ts/interfaces/ITelemetryData.ts
```diff
@@ -1,16 +1,39 @@
-interface ITelemetryEvent {
-  // The type of event we're sending, e.g. 'TYPERIGHTER_SUGGESTION_ACCEPTED' | 'TYPERIGHTER_MATCH_CREATED'
+export interface ITelemetryEvent {
+  /**
+   * The application sending the event
+   */
+  app?: string;
```
SHession, line 5: These properties are optional here but mandatory in the telemetry service definition.

src/ts/services/TelemetryService.ts
```diff
@@ -1,69 +1,55 @@
...
+    }
+
+    private throttledSendEvents = throttle(this.sendEvents, this.throttleDelay, { trailing: true, leading: false })
+
+    public addEvent(event: ITyperighterTelemetryEvent): void {
+      this.eventBuffer.push(event);
+      this.throttledSendEvents();
+    }
+
+    public flushEvents(): Promise<void> {
```
SHession, line 50: Is this going to be wired up in a subsequent PR?

SHession (commented):
This is excellent work! Reviewed the code and tested it locally and I can see its doing its job. Also I want to mention that the unit tests are awesome!
I've added a few comments worth considering.

src/ts/services/TelemetryService.ts
jonathonherbert: This anticipates being integrated in the telemetry service, where it's safe to have that opinion! It's a bit strange now, but it shouldn't be here for long.

src/ts/services/TelemetryService.ts
jonathonherbert: It's not really necessary in our page demo – we've tests for this, so I'm not too worried, but we could wire this up to a button or something to test if we need to.

SHession (approved):

PR #130, opened 2020-09-10
Add Typerighter match telemetry events

tjsilver:
## What does this change?

At the moment, we've no visibility of which matches are made in the document. 

This PR enqueues TYPERIGHTER_MATCH_FOUND events when matches are found in the document, via the [typerighter telemetry service](https://github.com/guardian/prosemirror-typerighter/pull/128). 

## How to test

1. Run this branch locally
2. Run `ts/add-telemetry-adapter-to-matcherservice` in Composer 
3. Use npm link
4. Check document using Typerighter and watch network requests. You should see requests that include 'TYPERIGHTER_MATCH_FOUND'.

We will receive events in the CELK stack for every Typerighter match found in a document.

We could have done this in one of two ways –

 1. Use the UI store subscription to listen for new matches, which will involved writing code to figure out what's new
 2. Pass the telemetryService into the MatcherService, and create the events as matches come in

Approach 1 involves more code, but gives a simpler public API. Approach 2 involves less code, but means our MatcherService now has to care about the TelemetryService.

We've opted for approach 2 now, as we'd like to ship this to users ASAP, with the option to refactor for approach 1 further down the line.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -66,14 +66,17 @@ if (editorElement && sidebarNode) {
     editorElement.getBoundingClientRect().height / 2 - menuHeight;
 
   const commands = createBoundCommands(view, getState);
+
+
+  const telemetryService = new TelemetryService("https://example.com") 
+  const typerighterTelemetryAdapter = new TyperighterTelemetryAdapter(telemetryService, "prosemirror-typerighter", "DEV");
+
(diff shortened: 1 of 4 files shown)


jonathonherbert (approved):
Looks good! Brilliant PR description 💯

PR #131, opened 2020-09-10
Add workflow file for github actions

jonathonherbert:
## What does this change?

Adds a Github workflow to build and release this project.

Releases are managed via semantic-release, via a [github action](https://github.com/marketplace/actions/action-for-semantic-release). This means automated, semantically versioned releases with nicely formatted release histories 🎉 

In order for semantic release to determine the correct version number, the project requires commits to be compliant with the [commitizen](https://github.com/commitizen) format.

This PR adds [git-cz](https://www.npmjs.com/package/git-cz) via a githook installed on `npm i` to provide an easy way to format commit messages. You can skip these! You just won't get a release.

## How to test

Does the github action run the correct code, and pass without errors?

Does this PR, when merged, correctly determine the new release number? (It shouldn't result in a release, as there aren't any changes to project code!)

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -7,7 +7,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ "*" ]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


jonathonherbert:
Closed in favour of #133.

PR #133, opened 2020-09-11
Add semantic release

jonathonherbert:
## What does this change?

Adds [semantic-release](https://github.com/semantic-release/semantic-release) via a workflow and [github action](https://github.com/marketplace/actions/action-for-semantic-release). This means automated, semantically versioned releases with nicely formatted release histories 🎉 

In order for semantic release to determine the correct version number, the project requires commits to be compliant with the [commitizen](https://github.com/commitizen-tools/commitizen) format.

This PR adds [git-cz](https://www.npmjs.com/package/git-cz) via a pre-commit githook installed on `npm i` to provide an easy way to format commit messages. You can skip these with `ctrl-c` – you just won't get a release.

Comments on this approach really welcome. I've used it in a personal project and found it very useful.

## How to test

Does the github action run the correct code, and pass without errors?

We can't see a release in master yet, but examine the actions run for the `beta` branch – you should see that a prerelease was published as a result of a `feat` commit. (I've changed that commit name for this branch to avoid triggering a build when this PR is merged, as for userland it's a no-op)

Does the readme accurately reflect the new process? Is there anything we should add/remove?

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -24,12 +24,15 @@ To test this plugin in applications that use it before publishing a release, use
 
 To update the project readme, edit the README.md in ./build and run `build:doc`. This runs Typedoc and appends the generated type information to the readme file, which is then published to the ./docs folder.
 
-## Publishing new versions
+## Commiting and publishing new versions
 
-To publish, you'll need to have an account on [NPM](https://www.npmjs.com/) and be a member of the [Guardian organisation](https://www.npmjs.com/org/guardian).
+This repository uses [semantic-release](https://github.com/semantic-release/semantic-release) to publish new versions of this package when PRs are merged to `master`, and prelease versions when code is pushed to `beta`.
 
-Before publishing, create a branch named and tagged with the [correct new version number](https://semver.org/), and merge it to master. To bump the version and create the tag, use npm via `npm version major | minor | patch`.
+Version numbers are determined by the commit history, and so to trigger a release you'll need to use the [commitizen](https://github.com/commitizen-tools/commitizen) format when writing commits.
 
-Once this is merged, you're ready to publish a new release. Running `npm publish` will build the application, generate its type declarations, and publish.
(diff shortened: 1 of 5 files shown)


build/README.md
```diff
@@ -24,12 +24,15 @@ To test this plugin in applications that use it before publishing a release, use
 
 To update the project readme, edit the README.md in ./build and run `build:doc`. This runs Typedoc and appends the generated type information to the readme file, which is then published to the ./docs folder.
 
-## Publishing new versions
+## Commiting and publishing new versions
```
tjsilver, line 27: Curious as to why we need two versions of the README?

SHession (approved):
I think this a great idea and I'm interested to see this working in the wild. I'm not sure how `Commitizen` will mesh with the way I like to commit little and often, but I am happy to try it.

build/README.md
jonathonherbert: Great question! The readme is autogenerated via `npm build:doc` by combining the readme in `./build` with generated docs. So a change in the `build` readme always has a commensurate change in `docs`.

It'd be nice to autogenerate documentation on a git action, for example when PRs are merged, to a) save devs seeing it in PRs and b) ensure the autogenerated documentation is never stale.

jonathonherbert:
@SHession it's worth noting that there are plenty of commit types that don't trigger a release – by default (which we're currently using), only [these types](https://semantic-release.gitbook.io/semantic-release/support/faq#how-can-i-change-the-type-of-commits-that-trigger-a-release) create releases. `chore` or `refactor` for little changes would be fine.

There's also nothing stopping you from using normal commit messages – they just won't produce releases.

I definitely think there's an open question around how hard it is to figure out which commit gets the versioned tag – for example, when developing a feature that requires lots of small changes. For example, in #126, the following commit history

```
Add testing utils
Create fixture for editor-level tests
Add test
Fix tests
```

Might now look like

```
test: add testing utils
chore: create fixture for editor-level tests
test: Add test
feat: preserve marks when issuing replacements
```

I think, as you say, we'll get a better sense of the friction in the wild.

PR #136, opened 2020-09-17
Check all categories by default

tjsilver:
## What does this change?
Currently  we manually specify all of the categories Typerighter must check on each request. As a result, the plugin asks for the current categories from the server before it's able to make requests to check documents. Related to this [Typerighter PR](https://github.com/guardian/typerighter/pull/70).

There are two problems with this:

1. there's a wait before you can hit 'Check document' when the plugin starts up
2. if the category list expands whilst the tool is in use, the plugin will miss categories. If it shrinks, the user sees an error.

This PR makes the change to remove category information from the request the plugin makes. 

## How to test
Run the branch locally (replacing `http://example.com` with the correct `telemetryService` link in `pages/index.ts`).
Check document and see that requests no longer contain categories.

By default, Typerighter checks incoming requests with every category.

### DIFF
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -19,7 +19,6 @@ interface IProps {
   setDebugState: (debug: boolean) => void;
   setRequestOnDocModified: (r: boolean) => void;
   requestMatchesForDocument: (requestId: string, categoryIds: string[]) => void;
-  fetchCategories: () => Promise<ICategory[]>;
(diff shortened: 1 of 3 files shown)


jonathonherbert (approved):
Tested locally, behaves as expected 👍

tjsilver:
[Check all categories by default](https://trello.com/c/XJmMnbn9/402-check-all-categories-by-default)

PR #137, opened 2020-09-17
Send more match data via telemetry

jonathonherbert:
## What does this change?

Add more match information, in tags, to the telemetry data we're sending from prosemirror-typerighter. This is to enable us to filter for matches marked as 'advisory', matches without suggestions, and matches with suggestions in aggregate in our telemetry data. There's also a `markAsCorrect` property we can use to identify 'correct' matches.

It also removes the `matchContext` metric, which is primarily used for detecting bugs and contains data we don't want to retain in telemetry.

## How to test

Run locally, and take an action that includes a match – accepting a suggestion, marking as correct, clicking in the sidebar. When telemetry events are sent a few seconds later, you should see that they contain the information above.

We can filter for advisory rules, matches without suggestions, 'correct' matches, and matches with suggestions in aggregate in our telemetry data.

### DIFF
--- a/src/ts/components/SidebarMatch.tsx
+++ b/src/ts/components/SidebarMatch.tsx
@@ -59,13 +59,7 @@ const SidebarMatch = ({
     e.preventDefault();
     e.stopPropagation();
 
-    telemetryAdapter?.sidebarMatchClicked({
-      documentUrl: document.URL,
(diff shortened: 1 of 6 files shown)


SHession (approved):
Reviewed the code and tested on my local machine, looks good!

PR #138, opened 2020-09-22
fix: 🐛 Fixes loading bar

SHession:
## What does this change?

This PR fixes the loading bar by accounting for the possibility of no categories.

This was an issue introduced by the ability to request a check with no categories, returning all default categories. In these circumstances the request blocks have no category ids associated, breaking how the remaining work percentage is calculated in the selector.

## How to test

Load the Prosemirror-Typerighter client, make a request, observe the loading bar. It should fill in relation to the matches found.

### DIFF
--- a/src/ts/state/selectors.ts
+++ b/src/ts/state/selectors.ts
@@ -71,9 +71,9 @@ export const selectPercentRemaining = <TMatch extends IMatch>(
     state.requestsInFlight
   ).reduce(
     ([totalWorkAcc, remainingWorkAcc], queryState) => {
-      const allWork = queryState.totalBlocks * queryState.categoryIds.length;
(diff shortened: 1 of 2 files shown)


jonathonherbert (approved):
Tested, works as expected, new test is 💯

jonathonherbert:
@SHession just to note that the reason each of these commits appear in the release notes generated by the merged PR is because each commit contains (inadvertently) a breaking changes flag, which I've just discovered [guarantees their inclusion](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/writer-opts.js#L45).

I've manually edited the release notes in question, but this shouldn't occur in future – and looking back, hasn't occurred for past releases.

It might be worth us raising something with git-cz re: the messaging, which I thought was also unclear!

PR #139, opened 2020-09-23
feat: 🎸 Add matcherType to telemetry 

tjsilver:
## What does this change?
At the moment, our rules don't include the name of their matcher. This will be useful in determining whether rules from different matchers have different telemetry  profiles. This is the sister PR to [this Typerighter PR](https://github.com/guardian/typerighter/pull/77).

## How to test
Run the branch locally and click `check document`. You will see after a few seconds that `matcherType` appears in the telemetry event request.

We will be able to determine whether rules from different matchers have different telemetry profiles.

### DIFF
--- a/src/ts/interfaces/IMatch.ts
+++ b/src/ts/interfaces/IMatch.ts
@@ -50,6 +50,7 @@ export type TMatchRequestErrorWithDefault = PartialBy<
 >;
 
 export interface IMatch<TSuggestion = ISuggestion> {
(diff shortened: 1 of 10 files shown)


tjsilver:
[Add matcher to telemetry tags](https://trello.com/c/6zoiMU8a/407-add-matcher-to-telemetry-tags)

jonathonherbert (approved):
Tested locally, works as expected 👍

PR #140, opened 2020-09-24
fix: 🐛 Adds the rule id to the feedback form

SHession:
## What does this change?

This PR adds the rule id to the "issue with result" feedback link. This should make it easier to identify the rule which has caused the issue.

## How to test

Submit an issue with a rule, look at the "Technical Context" and observe the rule id has been included.


jonathonherbert (approved):
Tested locally – thanks @SHession!

jonathonherbert:
Merged due to popular demand 😁

PR #141, opened 2020-09-24
Remove emojis from cz commits

jonathonherbert:
## What does this change?

`git-cz` is ... too fun by default. This dials back the fun to acceptable levels, that is to say, it is no longer fun.

## How to test

Take a look at the commit message below and adjust your monocle accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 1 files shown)


akash1810 (approved):
img

akash1810:
A more appropriate response...

img

PR #142, opened 2020-09-25
fix: 🐛 Remove height widget decorations – they're redundant

jonathonherbert:
## What does this change?

Height widget decorations are no longer needed since we moved our
[tooltip implementation to Popper.js](#112). Worse still – they cause a bug that prevents users selecting the left hand side of the annotated range.

This PR removes them, and the bug.

Before:

drag-ko

After:

drag-ok

## How to test

Running the demo locally, attempt to select the left hand side of a matched range.

The selection behaviour above should now be as expected.


SHession (approved):
Tested locally, LGTM!

PR #143, opened 2020-10-01
Add filter functionality, pt. 1 – add filter state and predicate to plugin state

jonathonherbert:
## What does this change?

Adds the ability to pass arbitrary filter state, and a filter predicate, to the plugin. Decorations and matches are filtered by the predicate when either the matches or the filter state changes.

This is part 1 of 2 – the next PR will add the UI changes that make altering this state possible.

## How to test

This is a no-op – the automated tests should pass.

The API looks sane, and the automated tests cover the cases we expect.

## Dev notes

This is a large PR! But, don't be alarmed. Most of the changes are related to an interface refactor that's largely ergonomic. 

There are two reasons it's so big –

- Adding another type parameter to the `IPluginState` interface meant a refactor to avoid having to pass `TMatch` and `TFilterState` to almost everything they touched was desirable – most of the time in functions that care about plugin state, we specific a generic plugin state that inherits from `IPluginState` and pass it transparently through
- For performance reasons, we need to cache the filtered state! My initial take recomputed the decorations on every keystroke, but with 200+ decorations to filter every time, this quickly becomes sluggish. This is in part because it's not possible to filter a `DecorationSet` directly – you must extract its decorations, filter them, and construct a new `DecorationSet`, and I suspect Prosemirror does more work than necessary when this happens as it's dealing with a whole new set of object identities.

I've added an – enum  👀  – to codify the match types, which were previous encoded in property names.

The filtering logic itself is kept in `/state/helpers.ts`.

### DIFF
--- a/src/ts/state/actions.ts
+++ b/src/ts/state/actions.ts
@@ -21,6 +21,7 @@ export const REMOVE_MATCH = "REMOVE_MATCH" as const;
 export const REMOVE_ALL_MATCHES = "REMOVE_ALL_MATCHES" as const;
 export const APPLY_NEW_DIRTY_RANGES = "HANDLE_NEW_DIRTY_RANGES" as const;
 export const SET_CONFIG_VALUE = "SET_CONFIG_VALUE" as const;
+export const SET_FILTER_STATE = "SET_FILTER_STATE" as const;
 
 /**
  * Action creators.
@@ -125,6 +126,18 @@ export const removeAllMatches = () => ({
 });
 export type ActionRemoveAllMatches = ReturnType<typeof removeAllMatches>;
 
+export const setFilterState = <TPluginState extends IPluginState>(
+  filterState: TPluginState["filterState"]
+) => ({
+  type: SET_FILTER_STATE,
+  payload: { filterState }
+});
+// tslint:disable-next-line:interface-over-type-literal
+export type ActionSetFilterState<TPluginState extends IPluginState> = {
+  type: typeof SET_FILTER_STATE;
+  payload: { filterState: TPluginState["filterState"] };
+};
+
 export type Action<TPluginState extends IPluginState> =
   | ActionNewHoverIdReceived
   | ActionNewHighlightIdReceived
@@ -137,4 +150,5 @@ export type Action<TPluginState extends IPluginState> =
   | ActionHandleNewDirtyRanges
   | ActionSetConfigValue
   | ActionRemoveMatch
-  | ActionRemoveAllMatches;
+  | ActionRemoveAllMatches
+  | ActionSetFilterState<TPluginState>;
--- a/src/ts/state/actions.ts
+++ b/src/ts/state/actions.ts
@@ -1,10 +1,9 @@
 import {
   IMatchRequestError,
   IMatcherResponse,
-  IRange,
-  IMatch
+  IRange
 } from "../interfaces/IMatch";
-import { IPluginConfig } from "./reducer";
+import { IPluginConfig, IPluginState } from "./reducer";
 
 /**
  * Action types.
@@ -49,16 +48,18 @@ export type ActionRequestMatchesForDocument = ReturnType<
   typeof requestMatchesForDocument
 >;
 
-export const requestMatchesSuccess = <TBlockMatches extends IMatch>(
-  response: IMatcherResponse<TBlockMatches>
+export const requestMatchesSuccess = <TPluginState extends IPluginState>(
+  response: IMatcherResponse<TPluginState["currentMatches"]>
 ) => ({
   type: REQUEST_SUCCESS,
   payload: { response }
 });
+// We must repeat ourselves here, as ReturnType doesn't play nicely with type parameters.
+// See https://github.com/microsoft/TypeScript/issues/26856
 // tslint:disable-next-line:interface-over-type-literal
-export type ActionRequestMatchesSuccess<TMatch extends IMatch> = {
+export type ActionRequestMatchesSuccess<TPluginState extends IPluginState> = {
   type: "REQUEST_SUCCESS";
-  payload: { response: IMatcherResponse<TMatch> };
+  payload: { response: IMatcherResponse<TPluginState["currentMatches"]> };
 };
 
 export const requestError = (matchRequestError: IMatchRequestError) => ({
@@ -73,22 +74,19 @@ export const requestMatchesComplete = (requestId: string) => ({
 });
 export type ActionRequestComplete = ReturnType<typeof requestMatchesComplete>;
 
-export const newHoverIdReceived = (
-  matchId: string | undefined
-) => ({
+export const newHoverIdReceived = (matchId: string | undefined) => ({
   type: NEW_HOVER_ID,
   payload: { matchId }
 });
 export type ActionNewHoverIdReceived = ReturnType<typeof newHoverIdReceived>;
 
-export const newHighlightIdReceived = (
-  matchId: string | undefined
-) => ({
+export const newHighlightIdReceived = (matchId: string | undefined) => ({
   type: NEW_HIGHLIGHT_ID,
   payload: { matchId }
 });
-export type ActionNewHighlightIdReceived = ReturnType<typeof newHighlightIdReceived>;
-
+export type ActionNewHighlightIdReceived = ReturnType<
+  typeof newHighlightIdReceived
+>;
 
 export const applyNewDirtiedRanges = (ranges: IRange[]) => ({
   type: APPLY_NEW_DIRTY_RANGES,
@@ -127,10 +125,10 @@ export const removeAllMatches = () => ({
 });
 export type ActionRemoveAllMatches = ReturnType<typeof removeAllMatches>;
 
-export type Action<TMatch extends IMatch> =
+export type Action<TPluginState extends IPluginState> =
   | ActionNewHoverIdReceived
   | ActionNewHighlightIdReceived
-  | ActionRequestMatchesSuccess<TMatch>
+  | ActionRequestMatchesSuccess<TPluginState>
   | ActionRequestMatchesForDirtyRanges
   | ActionRequestMatchesForDocument
   | ActionRequestError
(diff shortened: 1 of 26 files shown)


src/ts/state/actions.ts
```diff
@@ -127,10 +126,22 @@ export const removeAllMatches = () => ({
...
+  filterState: TPluginState["filterState"]
+) => ({
+  type: SET_FILTER_STATE,
+  payload: { filterState }
+});
+// tslint:disable-next-line:interface-over-type-literal
+export type ActionSetFilterState<TPluginState extends IPluginState> = {
+  type: typeof SET_FILTER_STATE;
+  payload: { filterState: TPluginState["filterState"] };
+};
```
sihil, line 139: Am I right to guess that this can't be `ReturnType<typeof setFilterState>` as `tsc` doesn't infer the type correctly?

src/ts/state/reducer.ts
```diff
@@ -76,8 +83,8 @@ export interface IBlockInFlight {
  * Handy when, for example, consumers know that parts of the document are
  * exempt from checks.
  */
-export type IIgnoreMatch = (match: IMatch) => boolean;
-export const includeAllMatches: IIgnoreMatch = () => false;
+export type IIgnoreMatchPredicate = (match: IMatch) => boolean;
```
sihil, line 86: Another type starting with `I`, albeit one you renamed rather than introduced...

src/ts/state/helpers.ts
```diff
@@ -20,3 +28,65 @@ export const addMatchesToState = <TMatch extends IMatch>(
...
 };
+
+/**
+ * Should we filter our decorations with the filterMatches predicate?
+ */
+export const shouldFilterDecorations = <TPluginState extends IPluginState>(
+  oldState: TPluginState,
+  newState: TPluginState,
+  filterMatches?: IFilterMatches<TPluginState["filterState"]>
+): filterMatches is IFilterMatches<TPluginState["filterState"]> => {
```
sihil, line 39: I feel like this function could do with a test to confirm that it does indeed correctly answer the question.

sihil (approved):
Some minor comments. questions and a suggestion for a test. Generally looks good though and actually quite a contained change once the type signature changes are eliminated.

Finally, as discussed, the application of hungarian notation to types/interfaces is superflous. Certainly worth stripping out at some stage but definitely a separate PR.

src/ts/utils/decoration.ts
```diff
@@ -3,20 +3,26 @@ import { Node } from "prosemirror-model";
...
-  unambiguousOpacity: string;
-  ambiguous: string;
-  ambiguousOpacity: string;
+export enum MatchType {
+  HAS_REPLACEMENT = "HAS_REPLACEMENT",
+  DEFAULT = "DEFAULT",
+  CORRECT = "CORRECT"
+}
+
+export interface IMatchTypeToColourMap {
```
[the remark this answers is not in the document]
jonathonherbert: It can and should be – but I'd suggest that's a change for another PR. Worth noting here, though, now that we have an enum value for rule types.

src/ts/state/reducer.ts
```diff
@@ -124,21 +139,41 @@ export interface IPluginState<TMatches extends IMatch = IMatch> {
...
 
+interface IInitialStateOpts<
+  TFilterState extends unknown,
+  TMatch extends IMatch
+> {
+  doc: Node;
+  matches: TMatch[];
+  ignoreMatch: IIgnoreMatchPredicate;
+  matchColours: IMatchTypeToColourMap;
+  filterOptions: IFilterOptions<TFilterState, TMatch> | undefined;
```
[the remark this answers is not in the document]
jonathonherbert: this: you must pass a value for `filterOptions`, even if it's undefined
`?`: you may optionally pass a value for `filterOptions`, but if you pass something, it must be defined

We use this here to allow us to transparently pass a possibly undefined value. If we didn't, we'd need to conditionally call the function with/without the argument depending on what the consumer code passed.

Having said that, the signature is a bit odd, eh? Maybe there's a better pattern here – I will have written this without thinking too hard to preserve the ergonomics in the calling code.

src/ts/state/helpers.ts
```diff
@@ -20,3 +28,65 @@ export const addMatchesToState = <TMatch extends IMatch>(
     decorations
   };
 };
+
+/**
+ * Should we filter our decorations with the filterMatches predicate?
+ */
+export const shouldFilterDecorations = <TPluginState extends IPluginState>(
```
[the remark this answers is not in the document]
jonathonherbert: Yeah, great suggestion. I'll rename.

src/ts/state/helpers.ts
```diff
@@ -20,3 +28,65 @@ export const addMatchesToState = <TMatch extends IMatch>(
...
+    !!filterMatches &&
+    (filterStateChanged || (matchesChanged && !noFilterApplied))
+  );
+};
+
+export const deriveFilteredDecorations = <TPluginState extends IPluginState>(
+  doc: Node,
+  newState: TPluginState,
+  filterMatches: IFilterMatches<
+    TPluginState["filterState"],
```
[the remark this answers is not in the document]
jonathonherbert: Yes, it is a bit dense! I've erred on the side of simplicity for the consumer at the expense of some internal complexity. If it's a bit much we can find another way – but it'll probably involve passing more types around in the calling code.

All of this type complexity is there to keep our state shape flexible to accommodate other filter types for third parties (or other filter shapes in other uses of Typerighter), and it's fair to say that it comes at a cost. We could reasonably draw a line and have an opinion here if we felt it was too much.

src/ts/state/reducer.ts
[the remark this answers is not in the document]
sihil: Super interesting. In that case you might like to supply a default value of undef in the function below as per the other fields.

src/ts/state/helpers.ts
```diff
@@ -20,3 +28,65 @@ export const addMatchesToState = <TMatch extends IMatch>(
...
+  );
+
+  const decorationsToAdd = createDecorationsForMatches(
+    matchesWithoutDecorations,
+    newState.config.matchColours
+  );
+
+  const decorationsToRemove = newState.decorations.find(
+    undefined,
+    undefined,
```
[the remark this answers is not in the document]
jonathonherbert: Haha, to be fair, this is just an ugly Prosemirror interface, I wouldn't go so far as to blame the language!

src/ts/state/actions.ts
jonathonherbert: You are! The function has a type parameter, so our usual ReturnType magic doesn't work here.

src/ts/state/reducer.ts
jonathonherbert: 😅 oh god, the sooner I get rid of these the better ...

src/ts/state/helpers.ts
jonathonherbert: You're right – we verify it via the plugin behaviour, but I had a nagging feeling that it'd be better to test in isolation. Let's do that.

src/ts/state/actions.ts
sihil: I thought as much. OOI how does it differ from `setConfigValue` which also seems to have types?

src/ts/state/reducer.ts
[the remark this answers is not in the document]
jonathonherbert: Ah yeah, that's what I'm after!

src/ts/state/reducer.ts
```diff
@@ -124,21 +139,41 @@ export interface IPluginState<TMatches extends IMatch = IMatch> {
...
-  const initialState: IPluginState<TMatch> = {
+export const createInitialState = <
+  TFilterState extends unknown,
+  TMatch extends IMatch
+>({
+  doc,
+  matches = [],
+  ignoreMatch = includeAllMatches,
+  matchColours = defaultMatchColours,
+  filterOptions
```
sihil, line 171: Think this needs `= undefined` for ease of use.

src/ts/state/actions.ts
jonathonherbert: The parameter types in setConfigValue aren't meant to be passed by the caller – they're used to infer argument types at the point the function is called.

`ConfigKey extends keyof IPluginConfig` is really saying 'any string literal in the union of the string keys of IPluginConfig', and we're using that type at the point of call to infer the `ConfigValue`.

This means we can only call `setConfigValue` with key-value pairs that are members of `IPluginConfig`, making it typesafe for consumers.

To demonstrate, you can see further down on L148 – we're not passing the` IPluginState` param into `setConfigValue`, but we do in e.g. `ActionRequestMatchesSuccess`.

src/ts/state/reducer.ts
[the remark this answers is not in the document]
jonathonherbert: Brief correction – I glanced at the definition above assuming it was part of the function definition, but it isn't – it's the definition of the options object that's passed to the function.

As a result there's no opportunity to supply a default in the interface, and supplying a default of `undefined` in the function implementation is a no-op, because the object is destructured when it's received. So this probably doesn't need a change as-is.

src/ts/state/helpers.ts
```diff
@@ -20,3 +28,65 @@ export const addMatchesToState = <TMatch extends IMatch>(
...
+  oldState: TPluginState,
+  newState: TPluginState,
+  filterMatches?: IFilterMatches<TPluginState["filterState"]>
+): filterMatches is IFilterMatches<TPluginState["filterState"]> => {
+  const matchesChanged = oldState.currentMatches !== newState.currentMatches;
+  const filterStateChanged = oldState.filterState !== newState.filterState;
+  const noFilterApplied = !oldState.filterState && !newState.filterState;
+
+  return (
+    !!filterMatches &&
```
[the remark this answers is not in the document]
jonathonherbert: Ahh, a single `!` is not such a misfortune, surely? Although to use two looks like carelessness.

... might be the only time I'd ever get to misquote Wilde in a code review

src/ts/state/reducer.ts
jonathonherbert: See comment above on why this shouldn't be necessary.

jonathonherbert:
Merging now, but we can always address any other concerns after the fact.

PR #144, opened 2020-10-01
Add filter functionality, pt. 2 – add filter UI

jonathonherbert:
NB: builds on #143 – please review that first!

## What does this change?

Building on #143, adds a filter UI to the sidebar results pane.

filter

Clicking on a coloured button toggles the filter state. You cannot turn all the match types off – once you're down to one, the button for the remaining type is disabled.

## How to test

The automated tests should pass.

Does the UI behave as expected? Are the matches ordered correctly?

The UI is becoming more feature-rich, and it feels like adding some integration tests might be useful fairly soon.

Users can filter by match type in a way that's convenient and intuitive.


tjsilver (approved):
Tested locally and works as expected.

PR #145, opened 2020-10-02
Allow for RHS sidebar UI display

SHession:
## What does this change?

This PR is designed to facilitate the display of Typerighter in the Composer RHS sidebar. This includes changing the style to accommodate the new display in addition to swapping the isActive functionality with the ability to clear returned matches.

This PR is designed to be tested alongside the currently open branch `ts-sh/move-typerighter-to-sidebar` in [flexible-content](https://github.com/guardian/flexible-content/tree/ts-sh/move-typerighter-to-sidebar).

## How to test

This can be tested by either setting up Composer locally and linking this version of the package using `npm link`. Alternatively a beta version can be released by merging into the `beta` branch and updating Composer to use that version of the package.

The Typerighter UI is able to render into the sidebar in a way that feels natural and intuitive. 

## Images

TRCLEAR


tjsilver:
[Move prosemirror-typerighter UI into the RHS menu of composer](https://trello.com/c/mfo1o1nS/416-move-prosemirror-typerighter-ui-into-the-rhs-menu-of-composer)

src/ts/interfaces/ITelemetryData.ts
```diff
@@ -40,6 +40,7 @@ export enum TYPERIGHTER_TELEMETRY_TYPE {
   TYPERIGHTER_MARK_AS_CORRECT = "TYPERIGHTER_MARK_AS_CORRECT",
   TYPERIGHTER_MATCH_FOUND = "TYPERIGHTER_MATCH_FOUND",
   TYPERIGHTER_CHECK_DOCUMENT = "TYPERIGHTER_CHECK_DOCUMENT",
+  TYPERIGHTER_CLEAR_DOCUMENT = "TYPERIGHTER_CLEAR_DOCUMENT",
```
jonathonherbert, line 43: V. minor – I'd recommend this (and other things further down, e.g. telemetry methods) be called 'clear matches' or similar – just because it's likely that in future we're clearing matches from things we might not call a document, e.g. headline, image caption etc.

src/ts/state/test/reducer.spec.ts
```diff
@@ -686,6 +687,72 @@ describe("Action handlers", () => {
...
+        tr,
+        newState,
+        removeAllMatches()
+      );
+
+      expect(newState.currentMatches).toEqual([]);
+      expect(newState.decorations).toEqual(state.decorations);
+    });
+    it("should remove error messages when they're present", () => {
+      const { state, tr } = createInitialData();
```
jonathonherbert, line 724: Ah, great to see a test for this case! 💯

jonathonherbert (approved):
Tested locally, looks great. Love to see the 'clear matches' button disabled when there are no matches and when a check is in progress 💯

It'd be good to address the comment above re: naming above if you agree, but it's not a blocker.

src/ts/interfaces/ITelemetryData.ts
SHession: I was aiming to be consistent with this naming as check and clear are antonyms in this context, so you would either be checking for matches or clearing matches. Therefore it made sense to me to name it this way. I'm happy to change it, but would we want to change the "TYPERIGHTER_CHECK_DOCUMENT" name too?

jonathonherbert: Haha, good point. Let's leave it for now – changing `CHECK_DOCUMENT`  would mean an inconsistency in our telemetry, and we know what we mean 😁

jonathonherbert (approved):

PR #146, opened 2020-10-02
Sort by matches

SHession:
## What does this change?

This PR adds the ability to sort Typerighter match results by importance (Red, ambder, green) or in order of appearance in the document. 

## How to test

Setup  the local Typerighter client, check the document, change the sort order of results. 

Users are able to see the match results in an order that is more appropriate for them.

## Images

TRSORT

### DIFF
--- a/src/ts/components/Results.tsx
+++ b/src/ts/components/Results.tsx
@@ -95,7 +95,7 @@ interface IProps {
               Results {hasMatches && <span>({currentMatches.length}) </span>}
             </span>
             <span className="Sidebar__header-sort">
-              Summary view 
+              Sort by colour 
               <Switch
                 size="small"
                 checked={sortAndGroup}
--- a/src/ts/components/Results.tsx
+++ b/src/ts/components/Results.tsx
@@ -97,7 +97,7 @@ interface IProps {
               Results {hasMatches && <span>({currentMatches.length}) </span>}
             </span>
             <span className="Sidebar__header-sort">
-              Sory By 
+              Sort by 
               <select className="Sidebar__header-sort-dropdown" value={matchSortBy} onChange={(event) => setMatchSortBy(event.target.value)}>
                 <option value={IMPORTANCE}>Importance</option>
                 <option value={APPEARANCE}>Appearance</option>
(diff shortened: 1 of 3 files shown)


src/ts/components/Results.tsx
```diff
@@ -34,8 +34,12 @@ interface IProps {
     getScrollOffset
   }: IProps) => {
 
+    const IMPORTANCE = "importance";
+    const APPEARANCE = "appearance";
```
jonathonherbert, line 38: Nit: I'd always put anything that's constant for each invocation outside of the function body. The performance difference in this case will be undetectable, so this is purely aesthetic and you're free to ignore!

src/ts/components/Results.tsx
```diff
@@ -93,7 +96,13 @@ interface IProps {
             <span>
               Results {hasMatches && <span>({currentMatches.length}) </span>}
             </span>
-
+            <span className="Sidebar__header-sort">
+              Sory By
```
jonathonherbert, line 100:
```suggestion
              Sort by 
```

src/ts/state/selectors.ts
```diff
@@ -130,4 +131,18 @@ export const selectRequestsInProgress = <TMatch extends IMatch>(
...
   state: IPluginState<TMatch>
-): boolean => state.config.isActive;
\ No newline at end of file
+): boolean => state.config.isActive;
+
+export const selectImportanceOrderedMatches = <TMatch extends IMatch>(
+  state: IPluginState<TMatch>
+): IMatch<ISuggestion>[] => sortBy(state.currentMatches, match => {
+  if(match.matchedText == match.replacement?.text){
+    return 2;
```
jonathonherbert, line 140: We can adjust this once #144 is in so both rely on a single source of truth.

jonathonherbert:
Sorry, just noticed this is in draft – please feel free to ignore the above!

SHession:
I've changed the dropdown to a toggle mode. This is in keeping with our new design intentions for the (filtering, sorting and grouping) 
image

jonathonherbert (approved):
This looks great. If we'd like to rebase @tjsilver's work offof this and get sorting and grouping out in one pass, I guess this PR is a no-op, but if we'd like to merge this, that'd be fine too. If the latter is the case, might be worth changing 'summary view' to 'sort by colour' or something, as idk if 'summary view' makes much sense if we're just sorting

PR #147, opened 2020-10-08
Add telemetry to summary toggle

SHession:
## What does this change?

This PR adds a telemetry event for when the "summary view" is toggled. This will be useful when determining if the toggle is an ability we want to keep.

BONUS: Adds the types to the telemetry events.

## How to test

Loads up the TR client, toggle the summary view ("sort by group"), observe the telemetry event in the outgoing traffic.

We are able to tell if the ability to switch between views is something users find helpful.


jennygrahamjones (approved):
Works as described 👍

PR #148, opened 2020-10-09
Group matches by rule ID

tjsilver:
## What does this change?
Before, users could only sort matches by importance (colour). Now the switch ('Summary view') has been adapted so that results are both sorted and matches with the same rule ID are grouped together under a dropdown. 'Summary view' is the default. Grouped matches are indicated by a dropdown arrow and number of matches in brackets. Once the user clicks the dropdown, they will see a short context snippet and the matched text in bold. They can then click on a match and scroll to the match in the document, as for single matches. Users can turn off summary view and the matches will all appear as single matches which are in document order. 

## How to test
Run the branch locally and experience the expected behaviour outlined above.

Telemetry indicates that users keep summary view switched on and that they are making use of the grouped match dropdowns.

We will need to monitor the responsiveness of the UI as a result of these modifications.

## Images
groupmatches_final

### DIFF
--- a/src/ts/components/SidebarMatchGroup.tsx
+++ b/src/ts/components/SidebarMatchGroup.tsx
@@ -33,9 +33,9 @@ const SidebarMatchGroup = ({
 }: IProps) => {
   
   const showGroupMatchSubset = () => (
-    <ul>
+    <ul className="Sidebar__list">
       {matchGroup.map(match => (
-        <li className="Sidebar__list-item" key={match.matchId}>
+        <li className="SidebarMatch__subset-list" key={`${match.ruleId}_${match.matchId}`}>
           <SidebarMatch
             matchColours={matchColours}
             match={match}
@@ -90,7 +90,7 @@ const SidebarMatchGroup = ({
               isGroup
               isSubset={false}
               showAllMatches={showGroupMatchSubset}
-              // numberOfGroupedMatches={matchGroup.length}
+              numberOfGroupedMatches={matchGroup.length}
             />
           </li>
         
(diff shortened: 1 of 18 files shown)


tjsilver:
[Group matches by ruleId in the sidebar](https://trello.com/c/oMxh9t0a/443-group-matches-by-ruleid-in-the-sidebar)

src/ts/components/SidebarMatchGroup.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const handleMouseEnter = () => {
+    indicateHighlight(firstMatch.matchId);
+  };
+
+  const handleMouseLeave = () => {
+    stopHighlight();
+  };
+
+  return (
+    <>
```
jonathonherbert, line 62: I don't think this fragment is needed anymore.

src/ts/utils/component.ts
```diff
@@ -0,0 +1,23 @@
+import { IMatch, TyperighterTelemetryAdapter } from "..";
+import { maybeGetDecorationElement } from "../utils/decoration";
+
+export const createScrollToRangeHandler = (match: IMatch, getScrollOffset: () => number, editorScrollElement: Element, telemetryAdapter?: TyperighterTelemetryAdapter) => (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
```
jonathonherbert, line 4: This file might benefit from some prettier love 💅

jonathonherbert (approved):
Couple of nits, this is brilliant work. 

Very, very minor styling – we don't use a 2px border as a delineation anywhere else in the sidebar, and the borders on the side are I think sufficient to make it clear we're looking at a snippet view, so I'd be tempted to reduce this to a single pixel. That's right, this review is actually suggesting a change of a single pixel. What am I. Even worse, I'm going to suggest it looks _cleaner_, ugh

Screenshot 2020-10-09 at 11 48 55

Screenshot 2020-10-09 at 11 49 03

(YMMV, please feel free to ignore)

PR #149, opened 2020-10-09
Add telemetry for filter toggles

jonathonherbert:
## What does this change?

Adds a telemetry event for the match filters, FILTER_STATE_CHANGED.

As a small multiplier bonus, exposes the telemetry event interface to the consumer, which I suspect will be handy elsewhere 😁 

## How to test

Toggle the match filter ... toggles. You should see events emitted by the telemetry service after the usual delay.

We can see when users toggle the match filters in our telemetry. 

## Dev notes

Included a minor refactor to DRY up our telemetry service a bit.

### DIFF
--- a/src/ts/components/FilterResults.tsx
+++ b/src/ts/components/FilterResults.tsx
@@ -1,6 +1,7 @@
-import React from "react";
+import React, { useContext } from "react";
 
 import { IMatch } from "..";
+import TelemetryContext from "../contexts/TelemetryContext";
 import {
   getColourForMatchType,
   getMatchType,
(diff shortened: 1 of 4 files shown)


SHession (approved):
This looks good and I've tested locally. 

`nit` I'm not sure the match types have great names, in particular  "HAS_REPLACEMENT" and "DEFAULT". I think these could be clearer.

jonathonherbert:
@SHession yes, maybe it'd be better to explicitly label them by their intent, rather than their derivation, as the intent is less likely to change over time.

PR #150, opened 2020-10-13
fix: Fixes the filter button number size

SHession:
## What does this change?

This PR fixes a minor issue with the filter options font size. This size is being overwritten in composer, causing it to display too large. This PR sets the size explicitly fixing the issue. 

## Images

before: 
image
image

after:
image
image


jennygrahamjones (approved):
Are any of the `$font-size-*` variables already used for the Sidebar styling appropriate for this? 

Either way, this looks good 👍

PR #151, opened 2020-10-13
Add a `getSkippedRanges` callback to allow consumers to ignore parts of text

jonathonherbert:
## What does this change?

The companion to this change in [Typerighter](https://github.com/guardian/typerighter/pull/85), this PR allows the plugin consumer to provide a function that maps from a Prosemirror `Node` to a series of ranges to ignore.

This is useful as we have a way of 'noting out' ranges in our CMS that remove the noted copy on publish. We'd like Typerighter to ignore these ranges when checking text.

## How to test

Ensure that your Typerighter server is up to date with the PR above, and run the client. It's been set up to ignore ranges marked as `code`:

- 'A sentence. and then another sentence' should trigger the the 'capitalise start of sentences' rule
- 'A sentence`.` and then another sentence' shouldn't.

The plugin consumer can provide ranges to ignore, and Typerighter ignores them correctly.


src/ts/utils/prosemirror.ts
```diff
@@ -217,9 +222,70 @@ export const applyPatchToTransaction = (
...
     const node = schema.text(patch.text, marks);
     return tr.insert(patch.from, node);
   }
   tr.delete(patch.from, patch.to);
 };
+
+/**
+ * Return the ranges covered by the given mark.
+ */
+export const findMarkPositions = (
```
jonathonherbert, line 235: Lifted from our CMS to provide a way of testing the getSkippedRanges function.

PR #152, opened 2020-10-13
fix: Fixes the scroll to match issue

SHession:
## What does this change?

This PR aims to fix the scroll-to match issue by tallying up the top offset value until the desired element is found. 

The bug was a result of using `offsetTop` regardless of where the match was found. If for example the match was within a list, the offset would only relative to the top of the list. This would result in the pane scrolling to the top regardless of where the match was located in the document.

There may be a better way of determining the element we want to calculate the offset relative to, using the `ProseMirror` class name seems fraught.

## How to test

Check a document with various matches, turn a set of matches into a list (or nested list), navigate to those match by using the sidebar. The panel should scroll to the correct location.

### DIFF
--- a/src/ts/utils/component.ts
+++ b/src/ts/utils/component.ts
@@ -16,7 +16,7 @@ export const createScrollToRangeHandler = (
     return;
   }
 
-  const scrollToYCoord = getMatchOffset(match.matchId) - getScrollOffset();
+  const scrollToYCoord = getMatchOffset(match.matchId, editorScrollElement) - getScrollOffset();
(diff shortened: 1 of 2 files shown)


jonathonherbert:
_Nice._ Re: className – would adding a `data-` attribute to the relevant node and using that as a selector be clearer?

tjsilver (commented):
I've tested this locally and it works!

jonathonherbert (approved):
Tested in Composer with the relevant markup, works as expected 👍

PR #153, opened 2020-10-14
Minor styling improvements 💅

jonathonherbert:
## What does this change?

Adds minor styling improvements: 

- Move opinions about sidebar borders outside of view component
- Ensure button's 'disabled' state is clear in consuming views
- Add shimmer to loading bar

## How to test

Fire up the UI. Does it correspond with the improvements above?

Users should detect a marginal increase in fanciness.

### DIFF
--- a/src/css/Controls.scss
+++ b/src/css/Controls.scss
@@ -23,12 +23,3 @@
   font-size: $font-size-small;
 }
 
-.LoadingBar {
-  position: absolute;
-  height: 3px;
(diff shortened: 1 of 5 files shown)


SHession (approved):
Tested changes locally, they looks like decent improvements. Discussed a slight change to the loading bar animation speed and made a commit for consideration. 

**EDIT**: The selector tests are now failing because of the change to the percentage selector method.

PR #159, opened 2020-11-02
feat: Adds debounce to match hover

SHession:
## What does this change?

This PR adds a small delay to the match popover to make navigating the document easier. This change was based on feedback provided from users that the popover we interfering with their regular work and would mean they would have to disable Typerighter to continue their usual work. 

## How to test

Navigate through a Typerighter annotated document, the popovers should not inhibit your ability to make changes.

Users are able to leave Typerighter enabled whilst editing the document.

This delay could cause a minor inconvenience to users and may require some fine tuning.

### DIFF
--- a/src/ts/components/MatchOverlay.tsx
+++ b/src/ts/components/MatchOverlay.tsx
@@ -1,5 +1,5 @@
 import Match from "./Match";
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect, useRef } from "react";
 import { IPluginState } from "../state/reducer";
 import { selectMatchByMatchId } from "../state/selectors";
 import { IMatch } from "../interfaces/IMatch";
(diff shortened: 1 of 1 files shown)


src/ts/components/MatchOverlay.tsx
```diff
@@ -40,6 +41,8 @@ const matchOverlay = <TPluginState extends IPluginState>({
     null
   );
   const [arrowElement, setArrowElement] = useState<HTMLDivElement | null>(null);
+  const [showMatch, setShowMatch] = useState<boolean>(false);
+  const debounceShowMatch = useRef(debounce(setShowMatch, 200))
```
jonathonherbert, line 45: Oh, this is really interesting. So `useRef` will only accept the first result of the `debounce` call, and use it for the entire lifecycle.

It's kindof weird to me that the result of subsequent calls to `debounce` get dropped on the floor, but this solution is extremely neat. Hooks, oh boy.

jonathonherbert (approved):
Nice and simple, behaves as expected and makes a real difference to popup flicker when traversing the document 👍

PR #160, opened 2020-11-03
feat: Handle skip ranges on the client

jonathonherbert:
## What's changed?

This PR adds logic to remove skipped ranges from blocks we send to our matcher service, and map returned matches through those ranges.

This was originally [implemented on the server](https://github.com/guardian/typerighter/pull/85), and in the accompanying PR there were a few notes as to why we'd made that choice. This PR effectively moves this logic to the client. 

This is because it's very difficult to debug problems with `skipRanges` logic that spans client and server. At the moment, we produce skipped ranges on the client, pass the block in its entirety to the server with the skippedRanges included, and rely on the server to do the right thing. But, if the server does the _wrong_ thing, it's difficult to understand where the problem lies –
- are the skippedRanges incorrectly placed?
- have the skippedRanges been incorrectly applied to the text?
- have the skippedRanges been incorrectly applied to the match?
- is the match incorrect?

Keeping all of the skipRanges logic on the client reduces the surface area of an investigation, by ensuring that the server only sees text. cc @sihil @aug24 

More detail about the problem below.

## Problem 

At the moment, when a match touches e.g. noted text, it's discarded when we try to add it to the document.

But we still check against noted text on the server. This can cause problems. E.g. for a rule that checks to see if the start of a sentence is capped up, where square brackets denote a noted range:

`... end of an example sentence[.] or is it?`

Typerighter will see

`... end of an example sentence. or is it?`

and we'll get a match suggesting `or -> Or` – but we want Typerighter to see

`... end of an example sentence or is it?`

## What's changed?

This PR removes the ranges returned by `getSkippedRanges` from blocks before they're sent to the matcher service, and maps matches returned from this service through those ranges to ensure they're placed correctly.

For example, given the phrase 
```
Example [noted ]text
|       |      |   |
0       8      15  19
```
 and a match for `text`:
- we remove the skipped ranges from the block text, so the snippet `[noted ]` is removed for the range `(8, 15)`
- the matcher service sees `Example text` where `text` covers the range `(8, 12)`
- the matcher service returns a match covering range `(8, 12)`
- the final match is mapped back through the skipped range on it to the correct position, `(16, 19)`

## How to test

The automated tests should be thorough and pass as expected.

Follow the instructions in [prosemirror-typerighter PR that introduced the `getSkippedRanges` function] (https://github.com/guardian/prosemirror-typerighter/pull/151) for an easy way to test manually.

## Definition of Success

Skipped ranges aren't seen by checkers, and their presence doesn't disturb the position of matches in the document.

### DIFF
--- a/src/ts/services/test/MatcherService.spec.ts
+++ b/src/ts/services/test/MatcherService.spec.ts
@@ -8,7 +8,10 @@ import { ITypeRighterResponse } from "../adapters/interfaces/ITyperighter";
 import { createBlockId } from "../../utils/block";
 import { IMatchRequestError } from "../../interfaces/IMatch";
 
-const createResponse = (strs: string[]): ITypeRighterResponse => ({
+const createResponse = (
+  strs: string[],
+  fromPos: number = 0
+): ITypeRighterResponse => ({
   requestId: "set-id",
   categoryIds: ["numberCat"],
   blocks: [
@@ -20,8 +23,8 @@ const createResponse = (strs: string[]): ITypeRighterResponse => ({
     }
   ],
   matches: strs.map(str => ({
-    fromPos: 0,
-    toPos: str.length,
+    fromPos,
+    toPos: fromPos + str.length,
     id: createBlockId(0, 0, 5),
     matchedText: str,
     message: "It's just a bunch of numbers, mate",
@@ -50,7 +53,8 @@ const block = {
   from: 0,
   to: 10,
   text: "1234567890",
-  id: "0-from:0-to:10"
+  id: "0-from:0-to:10",
+  skipRanges: []
 };
 
 const commands = {
@@ -60,8 +64,21 @@ const commands = {
 };
 
 const requestId = "set-id";
-
 const store = new Store();
+const endpoint = "http://typerighter-service-endpoint.rad";
+const createMatcherService = () =>
+  new MatcherService(store, commands as any, new TyperighterAdapter(endpoint));
+const getLastRequest = () => {
+  try {
+    const [, request] = fetchMock.lastCall()!;
+    const requestBody = JSON.parse(request!.body!.toString());
+    return requestBody;
+  } catch (e) {
+    throw new Error(
+      `Error parsing last request made to fetchMock: ${e.message}`
+    );
+  }
+};
 
 jest.mock("uuid", () => ({ v4: () => "id" }));
 
@@ -71,13 +88,9 @@ describe("MatcherService", () => {
     commands.applyMatcherResponse.mockReset();
   });
   it("should issue a fetch given a block, resolving with matches and broadcasting the correct event", done => {
-    const service = new MatcherService(
-      store,
-      commands as any,
-      new TyperighterAdapter("http://endpoint")
-    );
+    const service = createMatcherService();
     const response = createResponse(["1234567890"]);
-    fetchMock.post("http://endpoint/check", response);
+    fetchMock.post(`${endpoint}/check`, response);
 
     expect.assertions(1);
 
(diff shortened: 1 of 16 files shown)


src/ts/services/test/MatcherService.spec.ts
```diff
@@ -113,4 +121,53 @@ describe("MatcherService", () => {
...
+
+    it("On receipt of matches, map matches that succeed skipped ranges through those ranges to ensure they're correct", done => {
+      const service = createMatcherService();
+      // const response = createResponse(['F'], 2);
+      fetchMock.post(`${endpoint}/check`, 400);
+      service.requestFetchMatches();
+
+      store.emit("STORE_EVENT_NEW_MATCHES", requestId, [blockWithSkipRanges]);
+      setTimeout(() => {
+        // @todo
```
SHession, line 168: I assume there should be some test logic in here?

src/ts/utils/test/block.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        skipRanges
+      };
+      const newBlock = removeSkippedRanges(block);
+
+      expect(newBlock.text).toBe("DGJ");
+      expect(newBlock.from).toBe(40);
+      expect(newBlock.to).toBe(43);
+    });
+
+    it("should not care about order", () => {
```
SHession, line 77: It might be helpful to have some more awkward cases. For instance if ranges overlap or go beyond the boundaries of the block. Regardless of the behaviour it would be useful to have tests.

src/ts/utils/test/match.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+      const ruleMatch = getRuleMatch(10, 15);
+      const skippedRange = [{ from: 10, to: 15 }];
+      const mappedMatch = mapThroughSkippedRanges(ruleMatch, skippedRange);
+      expect(mappedMatch.from).toBe(16);
+      expect(mappedMatch.to).toBe(21);
+    });
+
+    it("should account for a range skipped within the given range, and extending beyond it", () => {
+      const ruleMatch = getRuleMatch(8, 12);
+      const skippedRange = [{ from: 8, to: 15 }];
```
SHession, line 38: Do we account for a case where a skipped range would put two characters next to each other which would previously have been separated? I.e. ABC**DEF**G -> ABCG -> matches on CG. How do those ranges get mapped?

src/ts/utils/test/range.spec.ts
```diff
@@ -338,4 +348,90 @@ describe("Range utils", () => {
       ]);
     });
   });
+  describe("mapAddedRange", () => {
+    it("should account for a range added before the given range", () => {
+      const incomingRange = { from: 10, to: 15 };
+      const addedRange = { from: 0, to: 5 };
```
SHession, line 354: `nit` I think the names incomingRange and addedRange don't mean much to me and make this harder to parse.

SHession (approved):
Tested locally and reviewed the code. Changes look sensible and the tests are very thorough. I've added a few small comments for your consideration.

src/ts/utils/block.ts
```diff
@@ -28,3 +39,41 @@ export const createMatchId = (
   to: number,
   index: number = 0
 ) => `${createBlockId(time, from, to)}--match-${index}`;
+
+/**
+ * Remove the given ranges from the block text, adjusting the block range accordingly.
+ */
```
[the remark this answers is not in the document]
jonathonherbert: I've added an example to the comment, which hopefully makes things clearer – lmk!

src/ts/utils/match.ts
```diff
@@ -0,0 +1,52 @@
...
+  match: TMatch,
+  blocks: IBlockWithSkippedRanges[]
+): TMatch => {
+  const maybeBlockForThisMatch = blocks.find(
+    block => match.from >= block.from && match.to <= block.to
+  );
+  if (!maybeBlockForThisMatch) {
+    return match;
+  }
+  const skipRangesForThisBlock = maybeBlockForThisMatch.skipRanges || []
```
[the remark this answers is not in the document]
jonathonherbert: Good spot, I changed an interface to make skipRanges non-optional for the `IBlockWithSkippedRanges` type, this is a hangover

src/ts/services/test/MatcherService.spec.ts
jonathonherbert: Added 👍

src/ts/utils/range.ts
```diff
@@ -159,27 +153,103 @@ export const expandRangeToParentBlockNode = (
...
+  const rangeBetweenRemovedStartAndIncomingStart = {
+    from: removedRange.from,
+    to: incomingRange.from
+  };
+  const intersection = getIntersection(
+    rangeBetweenRemovedStartAndIncomingStart,
+    removedRange
+  );
+  if (intersection) {
+    return intersection.to - intersection.from + 1;
```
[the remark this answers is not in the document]
jonathonherbert: I've updated this with comments – hopefully it makes a bit more sense!

src/ts/utils/test/range.spec.ts
jonathonherbert: I've changed `incomingRange` to `currentRange` – does that make more sense? Might bear some discussion.

src/ts/utils/test/match.spec.ts
jonathonherbert: This is I think accounted for in `should account for a range skipped within the given range` – we stretch the match across that range.

jonathonherbert: (What the client does this that, at the moment, is another story – we drop the match completely, as it touches a note. This is a naive approach that works on the assumption that any ranges touched by user-defined marks (notes, flags) have been reviewed, and we may well want to review it down the line.)

src/ts/utils/test/block.spec.ts
jonathonherbert: I've added tests for overlap and going beyond block range boundaries on both sides (thankfully without changing code!) – all good shouts.

SHession (approved):
LGTM! I suppose there will be an accompanying PR to revert the changes to the Typerighter server?

jonathonherbert:
@SHession absolutely, although this PR is compatible with the old interface, to, as skipRanges is optional – so I'll merge this first, and we can take care of removing the server side logic afterwards.

PR #161, opened 2020-11-10
fix: Remove matches when range edge touched

tjsilver:
## What does this change?

Adds the ability for a match and its associated decorations to be removed when the user clicks immediately to the left of it (this behaviour already existing for user clicking to the immediate right of a match).

## How to test

- Locally, check the text: 'We an hope to address it.' You should get a match for 'an'.
- and a `c` to correct 'an' to 'can'. The match should disappear from the document and Typerighter panel.

A match will disappear after the user has typed something immediately next to it, both its decoration in the document and its entry in the Typerighter panel.

### Before

remove_abutting_before

### After

remove_abutting

### DIFF
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -468,7 +468,7 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
   // and we need to ensure that the range includes the cursor position before it.
   newDecorations = removeDecorationsFromRanges(newDecorations, dirtiedRanges);
   const currentMatches = state.currentMatches.filter(
-    output => findOverlappingRangeIndex(output, dirtiedRanges, -1) !== -1
+    output => findOverlappingRangeIndex(output, dirtiedRanges, -1) === -1
   );
 
   return {
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -456,7 +456,7 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
   { payload: { ranges: dirtiedRanges } }: ActionHandleNewDirtyRanges
 ): TPluginState => {
   // Map our dirtied ranges through the current transaction, and append any new ranges it has dirtied.
-  const newDecorations = state.config.debug
+  let newDecorations = state.config.debug
     ? state.decorations.add(
         tr.doc,
         dirtiedRanges.map(range => createDebugDecorationFromRange(range))
@@ -466,7 +466,7 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
   // Remove any matches and associated decorations touched by the dirtied ranges from the doc
   // We are providing a from offset of -1 as the range provides a cursor position,
   // and we need to ensure that the range includes the cursor position before it.
-
+  newDecorations = removeDecorationsFromRanges(newDecorations, dirtiedRanges);
   const currentMatches = state.currentMatches.filter(output =>
     findOverlappingRangeIndex(output, dirtiedRanges, -1)
   );
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -464,11 +464,9 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
     : state.decorations;
 
   // Remove any matches and associated decorations touched by the dirtied ranges from the doc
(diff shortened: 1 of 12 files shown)


tjsilver:
Updating tests now.

SHession (approved):
Tested locally and reviewed the code, works well!

SHession:
It might be nice to add some unit tests for this if possible

src/ts/utils/decoration.ts
```diff
@@ -68,8 +68,8 @@ export const removeDecorationsFromRanges = (
   decorationSet: DecorationSet,
   ranges: IRange[],
   types = [DECORATION_MATCH, DECORATION_MATCH_HEIGHT_MARKER]
-) =>
-  ranges.reduce((acc, range) => {
+) =>{
+  return ranges.reduce((acc, range) => {
```
jonathonherbert, line 72: Is there a reason we've added a block here?

src/ts/state/reducer.ts
```diff
@@ -463,11 +463,12 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
...
-  // Remove any matches and associated decorations
-  // touched by the dirtied ranges from the doc
+  // Remove any matches and associated decorations touched by the dirtied ranges from the doc
+  // We are providing a from offset of -1 as the range provides a cursor position,
+  // and we need to ensure that the range includes the cursor position before it.
   newDecorations = removeDecorationsFromRanges(newDecorations, dirtiedRanges);
-  const currentMatches = state.currentMatches.filter(
-    output => findOverlappingRangeIndex(output, dirtiedRanges) === -1
+  const currentMatches = state.currentMatches.filter(output =>
+    findOverlappingRangeIndex(output, dirtiedRanges, -1)
```
jonathonherbert, line 471: This works – but possibly by coincidence! `findOverlappingRangeIndex` returns an index, which, if there's a single range dirtied, _happens_ to always to be zero (and falsy), and thus this filter behaviour works ... but are we sure that we'll always dirty a single range in a transaction?

My suspicion is that this will break if there are multiple ranges (e.g. on document cleanup (the wand icon in Composer), so I'd recommend being explicit with a comparison to an absent index (`!== -1`).

tjsilver: Oops yes I think the comparison got knocked out during the edit!

src/ts/state/reducer.ts
```diff
@@ -463,11 +463,12 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
...
 
-  // Remove any matches and associated decorations
-  // touched by the dirtied ranges from the doc
+  // Remove any matches and associated decorations touched by the dirtied ranges from the doc
+  // We are providing a from offset of -1 as the range provides a cursor position,
+  // and we need to ensure that the range includes the cursor position before it.
   newDecorations = removeDecorationsFromRanges(newDecorations, dirtiedRanges);
   const currentMatches = state.currentMatches.filter(
-    output => findOverlappingRangeIndex(output, dirtiedRanges) === -1
+    output => findOverlappingRangeIndex(output, dirtiedRanges, -1) === -1
```
tjsilver, line 471: As `!== -1` broke the functionality, have reinstated `===1`

tjsilver:
[Tooltips for ranges that span multiple lines always appear under the lower line on lhs, even when hovered on the upper line on the rhs](https://trello.com/c/JWUZ4Bbh/478-tooltips-for-ranges-that-span-multiple-lines-always-appear-under-the-lower-line-on-lhs-even-when-hovered-on-the-upper-line-on-th)

tjsilver:
[Remove matches when author touches range at edge of match](https://trello.com/c/RFPhBSsX/491-remove-matches-when-author-touches-range-at-edge-of-match)

src/ts/state/reducer.ts
jonathonherbert: oh oops, that's probably me misremembering what was there! (checks) – yep, my bad 😅

jonathonherbert (approved):
Paired on some of this, great set of changes 👍

jonathonherbert (approved):
WRT build issues, yep, well spotted. Still waiting on one CI (Travis is almost redundant with gh-actions, we only use it for codecoverage at this point – idk why it's been queued, possibly they're having problems), but it'd be odd if this failed and actions succeeded. Might be one for tomorrow.

PR #162, opened 2020-11-11
fix: popover placement issue

SHession:
## What does this change?

This PR handles the placement of popups on matches which cross multiple lines. This is important as previously some match popups would be impossible to reach due to their placement. 

This issue is fixed through the usage of `getClientRects` which provides a set of rectangles which an element covers, importantly with a separate rect per line. When the popup event is triggered on mouseover, the rect index is saved for reference. This is then used when determining the offsets of the popup and adjusting the offset X and Y based on the rect values.

## How to test

Run a document through TR, find a match which spans multiple lines, observe that it's possible to access a popup regardless of where the match is hovered over.

Users are always able to access a match popup and they are placed in a sensible position. 

It may be dangerous to second guess Popper's default behaviour and their may some unexpected issues a s result.

## Images

BEFORE
TRPOPUPOLD

AFTER
TRPOPUP

### DIFF
--- a/src/ts/components/MatchOverlay.tsx
+++ b/src/ts/components/MatchOverlay.tsx
@@ -10,7 +10,8 @@ import Store, { STORE_EVENT_NEW_STATE } from "../state/store";
 import { ApplySuggestionOptions } from "../commands";
 import { usePopper } from "react-popper";
 import { debounce } from "lodash"
-import { ModifierArguments, Options } from "@popperjs/core";
+import { ModifierArguments, Options, Placement} from "@popperjs/core";
+
 
 interface IProps<TPluginState extends IPluginState> {
   store: Store<TPluginState>;
@@ -69,7 +70,6 @@ const matchOverlay = <TPluginState extends IPluginState>({
     if (!currentMatchId) {
       debounceShowMatch.current.cancel();
       setShowMatch(false);
-      console.log(showMatch)
       return;
     }
     const matchElement = maybeGetDecorationElement(currentMatchId);
@@ -82,7 +82,7 @@ const matchOverlay = <TPluginState extends IPluginState>({
       name: 'hoverOverRect',
       enabled: true,
       phase: "read" as const,
-      fn: ({ state }: ModifierArguments<Options>) => {
+      fn: ({state}: ModifierArguments<Options>) => {
         if(currentRectIndex === undefined){
           return state;
         }
@@ -91,13 +91,36 @@ const matchOverlay = <TPluginState extends IPluginState>({
         if(offset){
           const hoverRect = rects[currentRectIndex];
           offset.x = hoverRect.x;
-          offset.y = offset.y - hoverRect.height * (rects.length - currentRectIndex - 1)
         }
         return state;
       },
     }),
     [currentRectIndex]
   );
+
+  // const offsetOptions = React.useMemo(() : [ number, number ] => {
+  //   if(referenceElement && currentRectIndex != undefined){
+  //     const rects = referenceElement?.getClientRects();
+  //     const hoverRect = rects[currentRectIndex];
+  //     const heightMultiplier = Math.max(0, rects.length - 1 - currentRectIndex);
+  //     const y =  hoverRect.height * heightMultiplier;
+
+  //     return [0, -y - 3]
+  //   }
+
+  //   return [0, -3]
+  // }, [currentRectIndex]);
+
+  const getOffsets = React.useMemo(() => ({ placement }: {placement: Placement}): [number?, number?] => {
+    if(referenceElement && currentRectIndex != undefined){
+      const rects = referenceElement?.getClientRects();
+      const hoverRect = rects[currentRectIndex];
+      const heightMultiplier = placement.indexOf("bottom") >= 0 ? Math.max(0, rects.length - 1 - currentRectIndex) : currentRectIndex;
+      const y =  -hoverRect.height * heightMultiplier;
+      return [0, y - 3]
+    }
+    return [0, -3]
+  }, [referenceElement, currentRectIndex])
 
   const { styles, attributes } = usePopper(referenceElement, popperElement, {
     placement: 'bottom-start',
@@ -109,7 +132,7 @@ const matchOverlay = <TPluginState extends IPluginState>({
       // If there's a gap, the tooltip library detects a `mouseleave` event
       // and closes the tooltip prematurely. We account for this with
       // padding on the tooltip container – see the styling for MatchWidget.
-      { name: "offset", options: { offset: [0, -3] } }
+      { name: "offset", options: { offset: getOffsets } }
     ]
   });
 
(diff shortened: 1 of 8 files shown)


src/ts/components/MatchOverlay.tsx
```diff
@@ -8,6 +8,8 @@ import Store, { STORE_EVENT_NEW_STATE } from "../state/store";
 import { ApplySuggestionOptions } from "../commands";
 import { usePopper } from "react-popper";
 import { debounce } from "lodash"
+import { Placement} from "@popperjs/core";
```
jonathonherbert, line 11: Nit: this file might need a `prettier` pass.

src/ts/components/MatchOverlay.tsx
```diff
@@ -69,6 +75,21 @@ const matchOverlay = <TPluginState extends IPluginState>({
     debounceShowMatch.current(true)
   }, [currentMatchId]);
 
+  const getOffsets = React.useMemo(() => ({ placement }: {placement: Placement}): [number | null | undefined, number | null | undefined] => {
+    const isTop = placement.indexOf("top") >= 0;
+    const isBottom = placement.indexOf("bottom") >= 0
+    if(referenceElement && currentRectIndex != undefined && (isTop || isBottom)){
```
jonathonherbert, line 81:
```suggestion
    if(referenceElement && currentRectIndex !== undefined && (isTop || isBottom)){
```

src/ts/components/MatchOverlay.tsx
```diff
@@ -69,6 +75,21 @@ const matchOverlay = <TPluginState extends IPluginState>({
...
+      const rects = referenceElement?.getClientRects();
+      const hoverRect = rects[currentRectIndex];
+      const lastRect = rects[rects.length-1]
+      const heightMultiplier = isBottom ? Math.max(0, rects.length - 1 - currentRectIndex) : currentRectIndex;
+      const y =  -hoverRect.height * heightMultiplier;
+      const x = hoverRect.left - lastRect.left
+      return [x, y - 3]
```
jonathonherbert, line 88: We could define a yOffset constant, describe what it's doing and why, and refer to that here, `3`'s a bit magic (it was probably a bit magic before, too 😅 )

SHession: Yeah it's a good point, I will make that change!

src/ts/state/reducer.ts
```diff
@@ -130,6 +130,9 @@ export interface IPluginState<
   // The id of the match the user is currently hovering over –
   // e.g. to display a tooltip.
   hoverId: string | undefined;
+  //The clientRectIndex of the hover mouseover event.
```
jonathonherbert, line 133: I'd suggest a bit more detail here to let nonplussed devs understand our intent, e.g:

```suggestion
  // The index of the clientRect closest to hover mouseover event.
  // Spans can contain multiple clientRects when they're broken across lines.
  // By indicating which clientRect we're closest to, we can position our match
  // popup next to the correct section of the span.
  // See https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects.
```

src/ts/utils/clientRect.ts
```diff
@@ -0,0 +1,26 @@
+const topBuffer = 2;
+const leftBuffer = 2;
+
+export const getClientRectIndex = (event: Event): number | undefined => {
```
jonathonherbert, line 4: I'd suggest adding a comment for this function to answer the question – get the client rect for what?

src/ts/components/MatchOverlay.tsx
```diff
@@ -69,6 +75,21 @@ const matchOverlay = <TPluginState extends IPluginState>({
     debounceShowMatch.current(true)
   }, [currentMatchId]);
 
+  const getOffsets = React.useMemo(() => ({ placement }: {placement: Placement}): [number | null | undefined, number | null | undefined] => {
+    const isTop = placement.indexOf("top") >= 0;
+    const isBottom = placement.indexOf("bottom") >= 0
+    if(referenceElement && currentRectIndex !== undefined && (isTop || isBottom)){
+      const rects = referenceElement?.getClientRects();
```
jonathonherbert, line 88: Any comments here to clarify what's going on what be really welcome! I understand the y-axis process, but the x-axis ... I guess if we're hovering over the same rect as the last rect, the offset is 0, else we offset by the hoverRect – ah, I see, because the lastRect is _always_ where the tooltip is positioned for a `bottomLeft` tooltip?

src/ts/components/MatchOverlay.tsx
```diff
@@ -69,16 +75,35 @@ const matchOverlay = <TPluginState extends IPluginState>({
     debounceShowMatch.current(true)
   }, [currentMatchId]);
 
+  const getOffsets = React.useMemo(() => ({ placement }: { placement: Placement }): [number | null | undefined, number | null | undefined] => {
+    // We provide a negative offset here to ensure there's an overlap
+    // between the decoration triggering the tooltip and the tooltip.
+    // If there's a gap, the tooltip library detects a `mouseleave` event
+    // and closes the tooltip prematurely. We account for this with
+    // padding on the tooltip container – see the styling for MatchWidget.
+    const yOffset = -3;
```
jonathonherbert, line 84: 👏

jonathonherbert (approved):
I can't break this! Yesss! 👏 

Tested (including mousing over to apply suggestions):
- single-line matches against window constraints on bottom and rhs
- multi-line matches, top and bottom, against those constraints

Not tested: 
- Matches that span many lines, e.g. whole paragraphs – but we shouldn't really be fielding these, anyway.

Minor comments re: readability, otherwise, ✨

src/ts/components/MatchOverlay.tsx
SHession: I've added some comments to this. It will only work at the moment due to the currently chosen placement. If we wanted to support other placements we would need to make some changes to how the X offset is calculated.

jonathonherbert: Ah brill, yes nice to make the dependency explicit, if only in text.

src/ts/state/reducer.ts
```diff
@@ -130,6 +130,13 @@ export interface IPluginState<
   // The id of the match the user is currently hovering over –
   // e.g. to display a tooltip.
   hoverId: string | undefined;
+  // The index of the clientRect closest to hover mouseover event.
+  // Spans can contain multiple clientRects when they're broken across lines.
+  // By indicating which clientRect we're closest to, we can position our match
+  // popup next to the correct section of the span.
+  // See https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects.
+  //This index used to know where to display a match popup
```
jonathonherbert, line 138:
```suggestion
```

src/ts/state/reducer.ts
```diff
@@ -130,6 +130,13 @@ export interface IPluginState<
   // The id of the match the user is currently hovering over –
   // e.g. to display a tooltip.
   hoverId: string | undefined;
+  // The index of the clientRect closest to hover mouseover event.
```
jonathonherbert, line 133:
```suggestion
  // The index of the clientRect closest to the hover mouseover event.
```
(my typo)

jonathonherbert (approved):

PR #163, opened 2020-11-11
chore: Rely on the PR name when determining version bump, not commits

jonathonherbert:
## What does this change?

This PR makes a change to our semantic-release workflow. Previously, we relied on our commit history to determine our version bump. With this change, we rely on the PR name when determining our version bump, not its commits. 

**Pros:** (please amend or suggest pros/cons below!)
- Githooks for commit linting are now unnecessary
- It's clear in CI that our sem-rel name is correct

**Cons:**
- Only one 'change' (to be reflected in the changelog) can be added per PR.

In order for this approach to work as described, we need to use the 'squash commit' option when we merge pull requests (currently our default).

I've added a PR lint step, which will let us know when our PR names are poorly formed in CI.

As adopted by [commercial-core](https://github.com/guardian/commercial-core).

## How to test

Difficult to test without merging, after which we should see –
- a linting step in our CI that fails with incorrectly formed PR names
- a release when we merge PRs that corresponds to the PR name

Less commit-to-commit friction for developers when working on this project.

## Dev notes

Thanks to @sndrs and @adamnfish for helping to refine this workflow – comments/crits v. welcome.


tjsilver (approved):
This is great, it makes a lot of sense to move the githook from the point of committing to the PR itself. 👍

SHession:
I think this a great idea, I have become fed up with the commit linting. 

A PR should be focused enough so that only one change item should be sufficient. If this isn't the case it should be split into separate PRs.

My only concern is using squash commits which I have tended to avoid in the past. This is mainly due to the loss of visible commit history, although if it makes the PR more easily found I could see this being beneficial. Is the reason we need them that we require a single commit in the sem-rel format and a conventional merge commit would not be formatted correctly?

SHession (approved):
I think this is definitely worth trying

jonathonherbert:
> My only concern is using squash commits which I have tended to avoid in the past. This is mainly due to the loss of visible commit history, although if it makes the PR more easily found I could see this being beneficial.

Yeah, I'm interested in the cons here.

In practice, I've only ever seen people treating PRs as atomic when managing e.g. reverts – with the caveat that it's always possible to trace the squashed commit back to the PR and the branch that it came from.

There's definitely a [debate](https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase) – given that we've used it as a default here I'm not too worried about merging this as-is, but I understand that rebasing offers a little more visibility and flexibility, and we can make a change in the future if we find ourselves constrained by our choices here.

PR #164, opened 2020-11-18
feat: add telemetry on match decoration click

jonathonherbert:
## What does this change?

Adds a telemetry event and a callback to enable us to post that event when the match decoration is clicked. Clicking on a match range will send that match's data to the telemetry client.

### But ... why?

We _think_, because of historical, workflow-related issues, users are selecting ranges and manually making changes, rather than using the green 'apply suggestion' button in the tooltip. This'll, I hope, help us find out.

You can of course make a cursor selection next to a word without clicking the surrounding span, but my hope is that this will pick up e.g. ~50% of the cases. When I went to select text, I found myself clicking within the span every time, using the bounding box as to orient the cursor, but others' behaviour may well be different.

Users may be annotating a broader range when they make their changes, though, and there's no defence against that.

Finally, there's also a question of whether clicks into match ranges constitute intent to edit. My assumption is that this is almost always the case.

## How to test

Run the app, scan a document, and click within a match decoration (the highlighted range covering. After the telemetry throttle threshold (10s), you should see a telemetry event signifying the click has occurred. It should include the match data.

I wrote an integration test, and in the process sadly found out that jsDOM cannot supply the necessary DOM API, so the Prosemirror top-level event handler throws. We'll need to integrate Jest with e.g. Puppeteer to run our  unit tests in proper browser environment.

We can pick up, most of the time, users clicking on matches.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -43,6 +43,12 @@ document.body.append(overlayNode);
 const isElementPartOfTyperighterUI = (element: HTMLElement) =>
   overlayNode.contains(element);
 
+const telemetryService = new TelemetryService("https://example.com");
+const typerighterTelemetryAdapter = new TyperighterTelemetryAdapter(
+  telemetryService,
+  "prosemirror-typerighter",
+  "DEV"
+);
 
 const { plugin: validatorPlugin, store, getState } = createTyperighterPlugin({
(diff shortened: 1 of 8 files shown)


tjsilver (approved):
Tested locally and works as described!

PR #165, opened 2020-11-26
fix: Memoize SidebarMatch to avoid rerendering as results arrive

jonathonherbert:
## What does this change?

Memoizes the `SidebarMatch` component, which is continually rerendered as our state changes.

This results in a significant improvement in our JS calc times. Timings taken when rechecking an article of 4,540 (to better isolate unneeded rerenders), averaged over 5 tries:

| Before | After |
|-|-|
| 6252ms | 4234ms |

## How to test

This is a no-op for features, but the impact is noticeable when using CPU throttling in e.g. Chrome.

### DIFF
--- a/src/ts/components/Markdown.tsx
+++ b/src/ts/components/Markdown.tsx
@@ -1,0 +1,12 @@
+import React from "react";
+import { getHtmlFromMarkdown } from "../utils/dom";
+
+const Markdown = ({ markdown }: { markdown: string }) => (
+  <div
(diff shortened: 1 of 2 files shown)


SHession (approved):
This looks like a smart thing to do. I haven't used memo before but looking at it, how did you decide which component to use it on? Why can't we use it on the SidebarMatch.tsx?

tjsilver (approved):

PR #172, opened 2021-02-03
chore: Update semantic-PRs.yml

jonathonherbert:
## What does this change?

Trigger our pull request linter on the `pull-request` event, not `pull-request-target`, b/c:

Screenshot 2021-02-03 at 09 11 27

... meaning it might possible to open a PR via a fork that ran code to e.g. pull out secrets. I haven't confirmed, but this tighter scope makes sense regardless.

Thanks @akash1810 for spotting!

## How to test

This is a no-op – our actions should continue to work as expected.

Fewer vulns in our build process.

### DIFF
--- a/.github/workflows/semantic-PRs.yml
+++ b/.github/workflows/semantic-PRs.yml
@@ -1,6 +1,6 @@
 name: 'Lint PR'
 on:
-    pull_request_target:
+    pull_request:
         types:
             - opened
             - edited


akash1810 (approved):
👍🏽 for future visitors, [here are GitHub's docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target).

PR #185, opened 2021-08-31
fix: Remove and re-add package-lock.json to ensure react-dom@17.0.2

jonathonherbert:
## What does this change?

Removes and re-adds the `package-lock.json` file from the project to fix a problem with inconsistent `react` and `react-dom` versions, as at the moment they're resolving to `17.0.2` and `16.13.1` respectively.

## How to test

Run `npm ls react` and `npm ls react-dom` in the project root. `17.0.2` should be the only version that resolves.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,12 +48,12 @@
       }
     },
     "@babel/helper-annotate-as-pure": {
-      "version": "7.7.4",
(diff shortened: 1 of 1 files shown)


rhystmills (approved):
Looks good to me!

PR #186, opened 2021-09-27
fix: Remove redundant tooltip styling

jonathonherbert:
## What does this change?

Removes redundant tooltip styling. As well as being unnecessary (due to switching to PopperJS #112), the styling is causing placement problems in consumer code at different zoom levels.

## How to test

You'll need to use `npm link` to use this with your consumer:

```bash
npm link # In prosemirror-typerighter dir
npm link @guardian/prosemirror-typerighter # in consumer project (e.g. Composer)
```

You should see accurate popup placement at non-100% zoom levels. E.g. at 90%:

| before | after |
| --- | --- |
| Screenshot 2021-09-27 at 09 48 14|Screenshot 2021-09-27 at 09 48 21 |

### DIFF
--- a/src/css/index.scss
+++ b/src/css/index.scss
@@ -105,12 +105,3 @@ hr {
 .TyperighterPlugin__decoration-container {
   position: absolute;
 }
-
-.TyperighterPlugin__tooltip-overlay {
-  // position: relative provides a new stacking context, which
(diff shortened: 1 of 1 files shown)


SHession (approved):
I tested main and this change locally on the demo client and was unable to replicate the issues. However I encountered no problems with removing this so happy to +1

PR #187, opened 2021-12-01
fix: Filter block queries to ensure zero-width blocks are not included in requests

jonathonherbert:
## What does this change?

Filters the block queries before they're added to our plugin state (and subsequently sent to our checking service) to cull any that are empty.

This should ensure that we don't send requests for ranges with no content.

For good measure, we remove any ranges that just contain whitespace.

## How to test

- The unit tests should pass, and you should be convinced they handle the scenario above.
- Try hitting removing all of the content in a document in a running demo instance, and hit 'check document'. You should not see a request go out. Contentful documents should issue requests as normal.

No requests for checks that contain empty ranges.

### DIFF
--- a/src/ts/components/Controls.tsx
+++ b/src/ts/components/Controls.tsx
@@ -11,7 +11,8 @@ import {
   selectHasAuthError,
   selectRequestsInProgress,
   selectHasMatches,
-  selectDocumentHasChanged
+  selectDocumentHasChanged,
+  selectDocumentIsEmpty
 } from "../state/selectors";
 import TelemetryContext from "../contexts/TelemetryContext";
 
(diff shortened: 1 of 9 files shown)


SHession (commented):
Changes make sense. I have tested locally and I can see that requests are not being sent for empty blocks.

I have a couple of thoughts: 

- Would it ever be desirable to check whitespace blocks, for example to warn users about their presence?
- The UI for checking an empty document is not ideal. I think it's unclear if the button is working.

Let me know what you think.

jonathonherbert:
Let's keep blocks that include whitespace, just in case.

Re: UI, agreed. Perhaps we disable the check document button with a tooltip on hover that explains why it's not available?

SHession:
> Re: UI, agreed. Perhaps we disable the check document button with a tooltip on hover that explains why it's not available?

Sounds like a good suggestion. Ana might have some ideas too.

jonathonherbert:
That's now done! Testing for emptiness is harder than it first appears, because empty of text !== empty of nodes – but hopefully the utility I've added looks well tested.

Rather than use a tooltip, which invites a whole bunch of questions over styling and consistency, I've compromised with a title tag for now, as I think this behaviour should be fairly clear to users if they ever do encounter it:

empty-doc

SHession (approved):
Changes look sensible. I have tested this locally and not been able to break it. 

It's a shame ProseMirror provides no way to check for any text content.

jonathonherbert:
> It's a shame ProseMirror provides no way to check for any text content.

Yeah, it's an odd one – `node.textContent` will give you all the text, but it's an expensive op on large docs and this will happen every keypress, so I guess we're stuck with my hack for now 😅

Issue #189, opened 2021-12-09
Matches in flight are not superceded by dirtied ranges when they are returned


jonathonherbert:
See video:

https://user-images.githubusercontent.com/7767575/145437529-b7e709d5-f541-4952-8e14-f9741a229a49.mov

jonathonherbert:
Addressed in #190.

PR #190, opened 2021-12-13
fix: Keep track of dirty ranges when requests are inflight so they supercede incoming matches

jonathonherbert:
## What does this change?

When Typerighter's operating in realtime mode (`requestMatchesOnDocModified = true`), it keeps track of dirtied ranges, and ensures that any ranges dirtied whilst a request for that range is in flight invalidate matches made in that request.

In non-realtime mode, it ... doesn't. See #189.

This change corrects that oversight, ensuring that dirtied ranges are tracked whilst requests are in flight. Once the request is complete, `dirtiedRanges` is reset, and ranges are no longer tracked.

## How to test

- The automated tests should pass, and you should be satisfied they test for the correct behaviour.
- Try to reproduce #189, perhaps by [manually increasing browser latency](https://umaar.com/dev-tips/66-network-throttling-profiles/#:~:text=Go%20to%20Settings%20%3E%20Throttling.,the%20Network%20Panel%20throttle%20options.). It should no longer be possible.

### DIFF
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -448,14 +448,15 @@ const handleNewDirtyRanges = <TPluginState extends IPluginState>(
     output => findOverlappingRangeIndex(output, dirtiedRanges) === -1
   );
 
+  const shouldPersistNewDirtyRanges = state.config.requestMatchesOnDocModified
+    || !!Object.keys(state.requestsInFlight).length;
+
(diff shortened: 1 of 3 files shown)


SHession (approved):
Changes make sense and tests look good. I have also attempted to replicate this issue locally and have been unsuccessful. LGTM!

Issue #191, opened 2021-12-13
The state property `requestsInFlight` is confusing 


jonathonherbert:
... as it's made up of a map from `requestId` to `IBlocksInFlightState`:

```ts
  requestsInFlight: {
    [requestId: string]: IBlocksInFlightState;
  };
```

Perhaps it'd be better as 

```ts
  requestsInFlight: {
    [requestId: string]: IRequestInFlightState;
  };
```

as `IBlocksInFlightState` really represents the state we carry as a result of a request.

jonathonherbert:
Addressed in #192.

PR #192, opened 2022-01-06
chore: Rename IBlocksInFlightState -> IRequestInFlight

jonathonherbert:
## What does this change?

Renames IBlocksInFlightState -> IRequestInFlight, to address the confusion described in #191.

## How to test

This is a no-op – the library should work as before, and the tests should pass.

### DIFF
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -57,10 +57,10 @@ import { ExpandRanges, IFilterOptions } from "../createTyperighterPlugin";
 import { getBlocksFromDocument, nodeContainsText } from "../utils/prosemirror";
 import { Node } from "prosemirror-model";
 import {
-  selectSingleBlockInFlightById,
-  selectBlockQueriesInFlightForSet,
+  selectSingleBlockInRequestInFlightById,
+  selectRequestInFlightById,
   selectMatchByMatchId,
-  selectBlockQueriesInFlightById
+  selectBlocksInFlightById as selectBlocksInFlightById
 } from "./selectors";
 import { Mapping } from "prosemirror-transform";
 import {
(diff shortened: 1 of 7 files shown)


src/ts/state/reducer.ts
```diff
@@ -563,7 +563,7 @@ const amendBlockQueriesInFlight = (
   if (!currentBlockQueriesInFlight) {
     return state.requestsInFlight;
   }
-  const newBlockQueriesInFlight: IBlocksInFlightState = {
+  const newBlockQueriesInFlight: IRequestInFlight = {
```
Fweddi, line 566: Just want to check - are we still satisfied with the variable name `newBlockQueriesInFlight` given we have renamed the typing?

Fweddi: Otherwise, LGTM!

Fweddi (approved):

src/ts/state/reducer.ts
jonathonherbert: Great catch – no we are not! I've amended that, and updated another small shibboleth left over from an early name change (blockQueriesInFlight should just be blocksInFlight).

src/ts/state/reducer.ts
```diff
@@ -57,10 +57,10 @@ import { ExpandRanges, IFilterOptions } from "../createTyperighterPlugin";
 import { getBlocksFromDocument, nodeContainsText } from "../utils/prosemirror";
 import { Node } from "prosemirror-model";
 import {
-  selectSingleBlockInFlightById,
-  selectBlockQueriesInFlightForSet,
+  selectSingleBlockInRequestInFlightById,
+  selectRequestInFlightById,
   selectMatchByMatchId,
-  selectBlockQueriesInFlightById
+  selectBlocksInFlightById as selectBlocksInFlightById
```
Fweddi, line 63: I can't tell if my eyes are deceiving me here, but the alias looks to be the same as the original?

jonathonherbert: Heh, well spotted – missed this artefact left by VSCode refactoring. Should've waited to merge! I'll follow up.

PR #193, opened 2022-01-06
chore: Remove unnecessary alias

jonathonherbert:
## What does this change?

ronseal.gif??

Thanks @Fweddi for the spot!


Fweddi (approved):
scooby-doo-unmasking-gif-except-its-the-same-person.gif

PR #195, opened 2022-01-19
fix: Add z-index to correct positioning of tooltips

jonathonherbert:
## What does this change?

Adds a z-index to our tooltip container, which ensures that tooltips are raised above sidebar content.

|Before|After|
|--|--|
| Screenshot 2022-01-19 at 13 04 24|Screenshot 2022-01-19 at 13 04 58

## How to test

Does this PR resolve the visual issues present in 'before', above?

### DIFF
--- a/src/css/index.scss
+++ b/src/css/index.scss
@@ -104,4 +104,6 @@ hr {
 
 .TyperighterPlugin__decoration-container {
   position: absolute;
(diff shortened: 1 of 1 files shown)


paperboyo (approved):
I believe what I see. I don’t need to understand.
nrr3iSj

PR #196, opened 2022-01-28
fix: Handle telemetry send errors

jonathonherbert:
## What does this change?

Handle telemetry send errors, which are currently unhandled and clutter our Sentry instrumentation.

## How to test

Introduce a network error when telemetry events are sent, perhaps by turning off the browsers network access by any means and running a document check.

You should see `warn `log messages, but no unhandled errors.

Screenshot 2022-01-28 at 08 49 38

Fewer unhandled errors cluttering our Sentry logs in our consumer CMS.


SHession:
It looks like there is quite a lot of changes in this PR outside of the description provided. I would recommend making the service change in the [new repo](https://github.com/guardian/editorial-tools-user-telemetry-service/pull/10) and I wonder if this error logs could be related to this [Composer PR](https://github.com/guardian/flexible-content/pull/4024).

jonathonherbert:
Yep, this is now in the wrong place after #198 – closing in favour of opening a PR in the relevant repository.

Some of this may well be caused by the problem you mention in that PR – regardless of the provenance of the errors, probably best to catch them w/in the client.

PR #202, opened 2022-03-03
chore: Pin 3rd party actions

jonathonherbert:
## What does this change?

Pins our semantic-release related GHAs, which are third party actions, to a particular commit hash, in line with the security advice for [third party actions.](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions)

## How to test

- Does the release portion of the github action still run as expected?
- Does the PR linting still run as expected?


twrichards (approved):
👏

PR #208, opened 2022-06-29
chore: Use vite

jonathonherbert:
**NB:** depends on #209 – please review that first 🙏 

## What does this change?

Swaps out rollup in favour of [vite](https://vitejs.dev/) to handle our dev and production builds. (Vite uses rollup under the hood for the build process.)

## Why?

Working with this library locally, the current, rollup-based workflow takes a very long time to reflect your changes in a development environment (between 10-20s) Once the build is ready, you must refresh your browser to see changes.

Vite's built to improve that experience:
- the build system uses ESM directly, rebuilding single files on the fly and serving them directly to the browser
- Vite will swap out the smallest portion of code possible when changes are made, statelessly altering your app on the fly to reflect changes to your app:

hmr

Vite doesn't provide on-the-fly typechecking out of the box, so we've added it by forking the server process and running `tsc` in watch mode in parallel via `run-p`.

## How to test

**NB:** As our dev-nginx config has had to change to enable Vite to talk to the site over websockets, I've taken the opportunity to change the local domain for consistency with other services. Update your  dev-nginx configuration with `dev-nginx setup-app ./nginx/nginx-mapping.yaml` before running these changes. The app now runs at `https://prosemirror-typerighter.local.dev-gutools.co.uk`. 

- Run the application locally and make a few changes to exercise the HMR and typechecking. Does the application behave as expected?
- Build `prosemirror-typerighter` locally, and use it in a consuming project, e.g. `flexible-content`, linking it via [yalc (or similar).](https://github.com/wclr/yalc). It should continue to work as normal. (Very happy to pair on this!)
- Todo: I'll issue a beta release for this build to verify this over npm, too.


src/css/Controls.scss
```diff
@@ -7,7 +7,7 @@
 
 .Controls__row + .Controls__row {
   padding-top: 0;
-  padding-bottom: $gutter-width / 2;
+  padding-bottom: calc($gutter-width / 2);
```
jonathonherbert, line 10: This maths op is deprecated in `dart-sass`, which Vite uses.

package.json
```diff
@@ -89,8 +67,16 @@
     "diff": "^4.0.2",
     "lodash": "^4.17.11",
     "prosemirror-changeset": "^2.1.2",
-    "prosemirror-tables": "^0.7.8",
+    "prosemirror-example-setup": "^1.1.2",
```
jonathonherbert, line 70: Our developer dependencies didn't really reflect what should be bundled – this change corrects that.

src/ts/components/Suggestion.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useContext } from "react";
-import jsDiff, { Change } from "diff";
+import { Change, diffChars } from "diff";
```
jonathonherbert, line 2: Mixing default and named exports in this way doesn't make too much sense given the usage below, and `jsDiff` doesn't have a default export (upsetting vite).

src/ts/services/adapters/TyperighterAdapter.ts
```diff
@@ -10,6 +10,7 @@ import {
   TRequestErrorCallback,
   TRequestCompleteCallback
 } from "../../interfaces/IMatcherAdapter";
+import { getErrorMessage } from "../../utils/error";
```
jonathonherbert, line 13: Necessitated by the difference in error types as a result of a TS upgrade.

PR #212, opened 2022-07-13
feat: Add a 'dev mode' flag to the view, and a few developer utilities

jonathonherbert:
## What does this change?

Adds a 'dev mode' parameter to the view function, which when `true` presents the user with two checkboxes that permit:

- Turning real-time checking on and off
- Adding annotations to ranges that have pending (orange) or in-flight (green) checks.

Pending checks are only really relevant in real time mode – they represent the ranges that have been 'dirtied' since the last check, and are due to be sent to Typerighter. In-flight checks have been sent to Typerighter and have not yet come back.

Here's how the UI looks:

Screenshot 2022-07-13 at 15 08 04

And here's the annotations at work, with an explanation:

pm-debug

I've defaulted dev mode to true in the sandbox – hopefully that makes sense.

## How to test

Fire up prosemirror-typerighter locally and have a play. Does it behave as expected? Do you think these things will be useful?

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -106,6 +106,7 @@ if (editorElement && sidebarNode) {
     editorScrollElement: editorElement,
     getScrollOffset,
     telemetryAdapter: typerighterTelemetryAdapter,
+    enableDevMode: true
   });
 
   // Handy debugging tools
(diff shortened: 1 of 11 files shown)


jonathonherbert:
@rebecca-thompson hm not removing those ranges is a bit lazy – thanks for noticing, let's remove them.

rebecca-thompson (approved):

PR #213, opened 2022-07-14
chore: Add instructions for local testing

jonathonherbert:
## What does this change?

Adds instructions for local testing. Cribbed from [prosemirror-elements.](https://github.com/guardian/prosemirror-elements)

## How to test

Do the instructions make sense? Do they work?


rebecca-thompson (commented):
In order for this to work I think you need to add the `yalc` command to `package.json`, eg https://github.com/guardian/prosemirror-elements/blob/main/package.json#L22

Is it worth removing the lines about using `npm link`? From my own experience running this locally works with yalc, but I get weird type errors when I try and use npm link. I haven't digged into why

rebecca-thompson (approved):

PR #214, opened 2022-07-21
feat: Add chunked adapter

jonathonherbert:
## What does this change?

Adds a TyperighterAdapter that handles a chunked response that returns [json-seq](https://en.wikipedia.org/wiki/JSON_streaming#Record_separator-delimited_JSON_2). The counterpart to [this PR](https://github.com/guardian/typerighter/pull/20) in Typerighter.

## Dev notes

Reading the stream was more involved than expected, because there is no guarantee that the chunks we read from the stream correspond to complete records – even if (as is explained in the PR above) the server is sending chunks that happen to correspond to complete records. As a result, we maintain a buffer and wait until we see a `RECORD_SEPARATOR` before flushing records to Typerighter.

That chunk-to-record behaviour isn't specified anywhere, and so it's best probably best not to rely on it, anyhow.

## How to test

- The automated tests should pass. The coverage should be good, as there are a few edge cases I wanted to guard against (and I discovered more when writing the tests.)
- Try running this branch against a version of typerighter Typerighter that supports chunked responses ([e.g. this branch.](https://github.com/guardian/typerighter/pull/20)) You should see a single request going out, and matches streaming into the document as usual.

### DIFF
--- a/src/ts/services/adapters/test/TyperighterChunkedAdapter.spec.ts
+++ b/src/ts/services/adapters/test/TyperighterChunkedAdapter.spec.ts
@@ -1,40 +1,123 @@
-import { noop, partial } from "lodash";
+import { noop } from "lodash";
 import { ReadableStream } from "stream/web";
-import { TextDecoder, TextEncoder } from "util";
-import {
+import { TextEncoder } from "util";
+import { IMatcherResponse } from "../../../interfaces/IMatch";
+import TyperighterChunkedAdapter, {
   readJsonSeqStream,
   RECORD_SEPARATOR
 } from "../TyperighterChunkedAdapter";
 
-describe("readJsonSeqStream", () => {
-  const textDecoder = new TextDecoder();
-  const textEncoder = new TextEncoder();
+const textEncoder = new TextEncoder();
 
-  const createJsonSeqArrFromRecords = (
-    records: Array<Record<string, unknown>>
-  ) => records.map(record => JSON.stringify(record) + RECORD_SEPARATOR);
+const createJsonSeqArrFromRecords = (records: any[]) =>
+  records.map(record => JSON.stringify(record) + RECORD_SEPARATOR);
 
-  const createStream = (records: string[]) => {
-    const stream = new ReadableStream({
-      start: controller => {
-        records.forEach(record =>
-          controller.enqueue(textEncoder.encode(record))
-        );
-        controller.close();
+const createStream = (records: string[]) =>
+  new ReadableStream({
+    start: controller => {
+      records.forEach(record => controller.enqueue(textEncoder.encode(record)));
+      controller.close();
+    }
+  });
+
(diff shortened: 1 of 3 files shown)


jonathonherbert:
Now ready for review.

src/ts/services/adapters/test/TyperighterChunkedAdapter.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  afterAll(() => {
+    global.fetch = localFetch;
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("on a successful response, it should call `onMatchesReceived` for each record, and finally call `onRequestComplete`", async () => {
```
rhystmills, line 40: Nit, but:
```suggestion
  it("should call `onMatchesReceived` for each record in a successful response, and finally call `onRequestComplete`", async () => {
```

src/ts/services/adapters/test/TyperighterChunkedAdapter.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+      onError,
+      onComplete
+    );
+
+    expect(onReceived).toHaveBeenCalledWith(input[0]);
+    expect(onReceived).toHaveBeenCalledWith(input[1]);
+    expect(onComplete).toHaveBeenCalled();
+  });
+
+  it("on an empty response, it should call `onRequestError`, and finally call `onRequestComplete`", async () => {
```
rhystmills, line 74: As above:
```
As above:
```suggestion
  it("should call `onRequestError` for an empty response, and finally call `onRequestComplete`", async () => {
```

src/ts/services/adapters/test/TyperighterChunkedAdapter.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+      onComplete
+    );
+
+    expect(onError.mock.calls[0][0]).toMatchObject({
+      message: "Typerighter did not send a response"
+    });
+    expect(onComplete).toHaveBeenCalled();
+  });
+
+  it("on an non-2XX response, it should call `onRequestError`, and finally call `onRequestComplete`", async () => {
```
rhystmills, line 93:
```suggestion
  it("it should call `onRequestError` for an non-2XX response, and finally call `onRequestComplete`", async () => {
```

rhystmills (approved):
Tested locally with the relevant [typerighter branch](https://github.com/guardian/typerighter/pull/20) and all is working as expected.

Some minor suggestions on the wording of test names but otherwise good to go.

(Should this be a major version change given that this expects `typerighter` to provide a different interface, i.e. the checkStream endpoint? My understanding of semver lore falls down here)

jonathonherbert:
Ace! Re: version, happy for this to be non-breaking – it's an opt in, we'll still continue to support the old endpoint.

PR #216, opened 2022-07-25
feat: Remove summary toggle

jonathonherbert:
## What does this change?

Removes the summary toggle. We know from our telemetry that it's no longer used frequently enough to warrant its existence:

Screenshot 2022-07-25 at 13 43 42

|Before|✨After✨|
|-|-|
|Screenshot 2022-07-25 at 13 44 59|Screenshot 2022-07-25 at 13 44 48|

## How to test

Is the toggle gone? Is any related code cleared out?


rhystmills (approved):
Code looks sensible, and running as intended locally 👍

PR #219, opened 2022-08-03
feat: Enable realtime checking without the user interacting with the contro…

rhystmills:
## What does this change?
Currently, real time checking is not kicking in until the 'Check Document' button is checked, which makes liveblog testing difficult.

With this PR - If 'requestMatchesOnDocModified' is true, request matches on initialisation of the view.

(Is this a bad idea?)

## How to test

Run `prosemirror-typerighter` locally according to the instructions in the readme (alongside `typerighter`. Tick 'Enable real-time checking'. Make changes to the document. Do checks take place? Are new matches checked?


jonathonherbert:
@rhystmills is this now addressed by #220?

PR #220, opened 2022-08-08
feat!: Provide matcher service, rather than expecting the consumer to provide one

rhystmills:
Co-authored by @jonathonherbert and @ParisaTork 

## What does this change?
This PR stops moves the responsibility for creating a MatcherService into `prosemirror-typerighter`, and no longer requires the consuming service to provide a MatcherService.

This simplifies the setup required by the consuming service. The work of providing a matcher service will no longer need to take place within `setupTyperighterView` in consumers, which means that we can run checking from `prosemirror-typerighter` without rendering the UI (e.g. in the liveblog).

This should be a no-op within `prosemirror-typerighter`, but will require some changes in the consumer because of the interface change.

## How to test
1. Run `prosemirror-typerighter` locally. Does the service run as expected? Do all tests pass?

### DIFF
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -103,9 +103,6 @@ export interface IPluginOptions<
    */
   onMatchDecorationClicked?: (match: TMatch) => void;
 
-  /**
-   * Expose useful utilities for developers.
-   */
   telemetryAdapter?: TyperighterTelemetryAdapter;
   
   adapter: IMatcherAdapter<TMatch>,
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -254,8 +254,9 @@ const createTyperighterPlugin = <TFilterState, TMatch extends IMatch>(
       const commands = createBoundCommands(view, plugin.getState);
       matcherService.setCommands(commands);
 
+      // Check the document eagerly on editor initialisation if 
+      // requestMatchesOnDocModified is enabled
       const pluginState = store.getState();
-
       if (pluginState){
         const { requestMatchesOnDocModified } = selectPluginConfig(pluginState)
         requestMatchesOnDocModified ?? requestMatchesForDocument(
--- a/src/ts/createTyperighterPlugin.ts
+++ b/src/ts/createTyperighterPlugin.ts
@@ -1,4 +1,4 @@
-import { applyNewDirtiedRanges } from "./state/actions";
+import { applyNewDirtiedRanges, requestMatchesForDocument } from "./state/actions";
 import {
   IPluginState,
   PROSEMIRROR_TYPERIGHTER_ACTION,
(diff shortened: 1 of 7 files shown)


src/ts/createTyperighterPlugin.ts
```diff
@@ -99,6 +102,13 @@ export interface IPluginOptions<
    * Called when a match decoration is clicked.
    */
   onMatchDecorationClicked?: (match: TMatch) => void;
+
+  /**
```
jonathonherbert, line 106: Think this comment may not apply.

src/ts/createView.tsx
```diff
@@ -65,6 +68,17 @@ const createView = <TPluginState extends IPluginState<MatchType[]>>({
   overlayNode.classList.add("TyperighterPlugin__tooltip-overlay");
   logger.info("Typerighter plugin starting");
 
+  const pluginState = store.getState();
+
+  if (pluginState){
```
jonathonherbert, line 73: I think that if we do want this (check the document eagerly on editor instantiation if this setting is enabled), we probably don't want to do it in `createView`, as this function won't fire as we expect in liveblogs.

Instead, perhaps worth adding to the `view` prop in the plugin – perhaps in another PR?

src/ts/services/MatcherService.ts
```diff
@@ -107,18 +114,27 @@ class MatcherService<TFilterState, TMatch extends IMatch> {
...
     const requestId = v4();
-    this.commands.requestMatchesForDirtyRanges(
+    commands.requestMatchesForDirtyRanges(
       requestId,
       this.getCurrentCategories().map(_ => _.id)
     );
   }
 
+  /**
+   * Provide the matcher service with commands, which required an editor
```
jonathonherbert, line 132:
```suggestion
   * Provide the matcher service with commands, which must be bound to an `EditorView` instance, and so cannot be provided until the Typerighter plugin is instantiated by an `EditorView`.
```

jonathonherbert (commented):
Looks good! One blocking comment re: eager checking.

src/ts/createTyperighterPlugin.ts
rhystmills: Whoops - yes I think that probably applied to `enableDevMode`

src/ts/createView.tsx
rhystmills: I've moved to the `view` prop for now, but I will take it out of the PR if you feel strongly!

src/ts/services/MatcherService.ts
rhystmills: Now updated.

src/ts/createView.tsx
jonathonherbert: No problem – I don't think it's working at present, as we need to call the bound commands, and at the moment we're importing an unbound one. I've pushed a fix in [b5b77f6](https://github.com/guardian/prosemirror-typerighter/pull/220/commits/b5b77f6a5ecf0444c0870c8c0f1a65c029a787a9) – I'll approve this, might be worth a quick look to verify that's correct before merging 👍

jonathonherbert (approved):

PR #222, opened 2022-08-11
fix: Fix redundant console warn

rhystmills:
A redundant console warn slipped through.

This PR makes it do something more useful.

### DIFF
--- a/src/ts/services/MatcherService.ts
+++ b/src/ts/services/MatcherService.ts
@@ -43,7 +43,6 @@ class MatcherService<TFilterState, TMatch extends IMatch> {
     });
   }
   public getCommands = () => {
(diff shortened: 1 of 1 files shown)


jonathonherbert:
Just reminding myself – did we want to warn if `this.commands` was not there? IIRC it's still possible for this to occur if things are misconfigured (not that we expect this in our CMS 😇 )

rhystmills:
> Just reminding myself – did we want to warn if `this.commands` was not there? IIRC it's still possible for this to occur if things are misconfigured (not that we expect this in our CMS 😇 )

Ah - you may be right. Seems harmless to warn in case it isn't defined.

ParisaTork (approved):

PR #223, opened 2022-08-12
chore: Trigger snyk

kenoir:
Snyk integration. This PR tests whether or not integration with a snyk GHA will work out of the box. If it passes and has the correct org name it can be merged.


kenoir:
The [error](https://github.com/guardian/prosemirror-typerighter/runs/7805952746?check_suite_focus=true#step:6:29) indicates `esbuild-android-64 was not found in package-lock.json`. It is [referenced as an `optionalDependency`](https://github.com/guardian/prosemirror-typerighter/blob/main/package-lock.json#L11346) though??

mchv:
@jonathonherbert Could you look to integrate this please?

jonathonherbert:
@kenoir @jfsoul – looks like this is now working. No change on our side, so looks like Snyk has solved. Thanks for raising the PR! Should we close this, or merge?

PR #224, opened 2022-08-22
fix: Tweak to center info icon

jonathonherbert:
## What does this change?

A small tweak to keep the information icon added in #217 centered in our consuming CMS.

|Before|After|
|-|-|
|Screenshot 2022-08-22 at 14 12 56|Screenshot 2022-08-22 at 14 37 10|

Quite subtle – I've adjusted the position of the icon so it should be dead centre, and removed the gray border above the coloured banner, which in my eyes altered the perception of central-ness for the icon.

## How to test

Take a look at the screenshots, or run locally. (Test in Composer is possible via the instructions in the readme.) Do you get a sense of ... geometric harmony?

### DIFF
--- a/src/ts/components/Tooltip.tsx
+++ b/src/ts/components/Tooltip.tsx
@@ -54,13 +54,11 @@ export const getPopperConfig = (
 };
 
 const tooltipIconView = css`
-    padding: 2px;
-    height: 20px;
+    height: 18px;
     margin-right: 4px;
   svg {
     height: 18px;
     width: 18px;
-    margin: -1px;
   }
   :hover {
     cursor: pointer;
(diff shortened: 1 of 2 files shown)


paperboyo:
💎

src/ts/components/Tooltip.tsx
```diff
@@ -54,13 +54,11 @@ export const getPopperConfig = (
 };
 
 const tooltipIconView = css`
-    padding: 2px;
-    height: 20px;
+    height: 18px;
```
dskamiotis, line 57: I think the padding had n impact in the `i` being very slightly off centre.
This seems to have done the trick :)

dskamiotis (approved):
Tested Locally - running Typeright, Prosemirror-typerighter and checking in the sandbox: `https://prosemirror-typerighter.local.dev-gutools.co.uk/`

PR #169, opened 2021-01-21
feat: UI overhaul

jonathonherbert:
## What does this change?

A UI overhaul, including
- more accessible matches
- headers in the sidebar to separate and explain match types

They're based on [this Figma file](https://www.figma.com/file/5RM0QzmQe8vCcVY9cX4KLL/Accessibility?node-id=619%3A11119), with one conscious alteration – I've darkened the orange match colour to `300` to allow the icon colour to remain white in all cases.

## How to test

These changes mostly concern the UI – there are no new unit tests. See success criteria.

Here's how it looks downstream in Composer, the Guardian's CMS:
 
Screenshot 2022-07-25 at 10 44 40

To test yourself in Composer, follow the instructions introduced in [this PR.](https://github.com/guardian/prosemirror-typerighter/pull/213)

## How can we measure success?

Users find the new UI more accessible, the filter options are more clearly signposted, and it's clearer what each match colour means.

PR #225, opened 2022-08-24
fix: Fix styling when correct and default matches span multiple lines

jonathonherbert:
## What does this change?

Fixes an issue where as a result of changes in #169, match styling would not apply correctly when matches were broken across lines.

|Before|After|
|-|-|
Screenshot 2022-08-24 at 09 43 19|Screenshot 2022-08-24 at 10 04 04|

Previously, we were using the `:after` pseudo element to achieve this effect with a background image. This doesn't play nicely with the [multiple border-boxes that constitute a broken span](https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects), but [border-image (TIL!)](https://developer.mozilla.org/en-US/docs/Web/CSS/border-image) does.

One caveat – the green line is some subpixel-width fatter than before, I think as a result of being exactly rather than approximately 3px. I think it's still clear. Sadly, I haven't found a way to make match the previous styling exactly, and the next pixel down looks too thin to me: happy to revise. Here's the thinner line:

Screenshot 2022-08-24 at 09 42 25

Tested in Chrome and Firefox.

## How to test

Run locally in the sandbox, or use the instructions in ['Testing locally in applications ...'](https://github.com/guardian/prosemirror-typerighter#testing-locally-in-applications-that-use-prosemirror-typerighter) to test in Composer. Do match decorations now behave themselves when split over multiple lines?

### DIFF
--- a/src/css/Sidebar.scss
+++ b/src/css/Sidebar.scss
@@ -32,7 +32,9 @@
 .Sidebar__header-container {
   position: relative;
   padding: $gutter-width;
-  border-bottom: 1px solid $sidebar-border-color;
+  & + & {
+    border-top: 1px solid $sidebar-border-color;
+  }
 }
 
 .Sidebar__header-bottom {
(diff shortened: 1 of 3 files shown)


rebecca-thompson (approved):
I have tested this locally, can replicate the bug and see this pr fix it

PR #229, opened 2022-09-12
chore: Add dev tools as a package dependency

jonathonherbert:
## What does this change?

Adds prosemirror-dev-tools as a package dependency. Previously, we were sourcing this from a script in the html.

In additional to restoring the cute owl and enabling development without an internet connection, this should remove a problem where devs saw an error in the console:

```
Uncaught RangeError: Content hole not allowed in a mark or leaf node spec
    at o.renderStructure (prosemirror-dev-tools.min.js:1:365845)
    at o.serializeMark (prosemirror-dev-tools.min.js:1:365050)
    at prosemirror-dev-tools.min.js:1:364504
    at Fragment2.forEach (fragment.js:175:7)
    at o.serializeFragment (prosemirror-dev-tools.min.js:1:364234)
    at o.renderStructure (prosemirror-dev-tools.min.js:1:365952)
    at o.serializeNode (prosemirror-dev-tools.min.js:1:364671)
    at prosemirror-dev-tools.min.js:1:364575
    at Fragment2.forEach (fragment.js:175:7)
    at o.serializeFragment (prosemirror-dev-tools.min.js:1:364234)
```

## How to test

Run the project locally. Do the dev tools behave as expected? Is the owl back?


ParisaTork (approved):

ParisaTork:
Can confirm the cute owl is indeed back and that the changes have been successful:

Imgur

Imgur

PR #233, opened 2022-09-23
feat: Read percentage complete from Typerighter API response when available

rebecca-thompson:

### DIFF
--- a/src/ts/state/reducer.ts
+++ b/src/ts/state/reducer.ts
@@ -157,6 +157,8 @@ export interface IPluginState<
   docChangedSinceCheck: boolean;
   docIsEmpty: boolean;
   typerighterEnabled: boolean;
+  // the percentage of blocks that have been processed by the matcher service
+  percentageRequestComplete?: number;
 }
 
 // The transaction meta key that namespaces our actions.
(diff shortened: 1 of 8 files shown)


src/ts/state/reducer.ts
```diff
@@ -157,6 +157,8 @@ export interface IPluginState<
   docChangedSinceCheck: boolean;
   docIsEmpty: boolean;
   typerighterEnabled: boolean;
+  // the percentage of blocks that have been processed by the matcher service
```
jonathonherbert, line 160: Pedantically, that isn't quite right – at the moment, it happens to be the percentage of the number of jobs that have been completed, but this is an implementation detail that might change. The variable name does a good job here, which makes the below suggestion a bit redundant, but more accurate I think – alternatively the comment is perhaps not needed.

```suggestion
  // the percentage of the request that has been processed by the matcher service
```

jonathonherbert (approved):
As with the review of https://github.com/guardian/typerighter/pull/204, I haven't been able to exercise this, but code looks good 👍

jonathonherbert:
Tested locally, works as expected.

PR #235, opened 2022-10-03
feat!: Separate out view functions for sidebar and overlay

jonathonherbert:
## What does this change?

Separates out view functions for sidebar and overlay views, and has them return unmount functions.

This addresses two oversights:
- we may want more than one instance on the page at once, and we may need to clean up after it is gone
- we may want to instantiate an overlay without a sidebar (in the Guardian's case, in liveblogs.)

## How to test

This is a no-op – the application should run as before. You should see two view functions, rather than one, exported from the project root.


rebecca-thompson (approved):
I have tested this locally and it works as expected. This is a sensible change. My only (optional) suggestion is to split the views into their own files with file names reflecting the component names

jonathonherbert:
Good shout! I've also moved them into Components, just to keep the UI bits in one place.

PR #236, opened 2022-10-04
feat!: Export getState

jonathonherbert:
## What does this change?

Exports a `getState` function that will return the Typerighter plugin state, given a general `EditorState`, alongside the relevant selectors to get hold of particular bits of plugin state.

This simplifies a few signatures that we export, and makes getting hold of the plugin state from an editor instance simpler in consuming code.

This is a breaking change, as it alters the signature of `boundCommands`.

## How to test

- The automated tests should pass.
- The use case should make sense!

Consuming code has an easier time accessing an editor's Typerighter state.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -53,7 +53,6 @@ const typerighterTelemetryAdapter = new TyperighterTelemetryAdapter(
 const {
   plugin: validatorPlugin,
   store,
-  getState,
   matcherService
 } = createTyperighterPlugin({
   isElementPartOfTyperighterUI,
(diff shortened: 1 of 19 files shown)


rebecca-thompson (approved):
I think this is fine for now. In future, when more of the composer codebase is converted to React, it would be good to revisit this decision and hopefully replace exposing internal plugin state with React Context as a means of sharing typerighter settings across editor boundaries

jonathonherbert:
@rebecca-thompson I think this PR will be necessary for using APIs like Context (which IIUC is not able to solve the problem of moving state into a React context from outside of it), although Context might be a good way to solve for bits of the Typerighter UI once Typerighter's state is available inside of React – let's chat out of band to make sure my assumptions are correct here.

PR #238, opened 2022-10-14
fix: Clean up unmounting React components correctly

jonathonherbert:
## What does this change?

There's a bug in #235 – the overlay does not clean up correctly, as it returns the wrong function to useEffect.

This PR cleans it up.

## How to test

- The automated test should pass.

I've added a test for the other component that (correctly) cleans up its subscription, and bumped a few jest dependencies to remove ts-jest warnings and keep us up to date.

No errors on unmount in consumer code.


dskamiotis (approved):
All automated tests are passing successfully 🥳

PR #242, opened 2022-10-27
feat: add telemetry event for check range event

rebecca-thompson:
## What does this change?
This adds a telemetry event `TYPERIGHTER_CHECK_RANGE` for when a range is checked by Typerighter. Prior to this, we only had a telemetry event when the entire document is checked. When realtime checking is enabled, the entire document is only checked once when the editor is initialised. Any subsequent checks are made on a range, not the whole document and we would like to have telemetry to know when this happens.

## How to test
Run prosemirror-typerighter locally and enable realtime checking. Let Typerighter run once, then add some text to the document so Typerighter runs a second time. In the network tab you should see a telemetry event sent (filter network traffic by 'event') with a type of `TYPERIGHTER_CHECK_RANGE`


ParisaTork:
Tested locally and can confirm the changes have been successful.

A screenshot showing the presence of TYPERIGHTER_CHECK_RANGE:
TYPERIGHTER_CHECK_RANGE

ParisaTork (approved):

PR #243, opened 2022-10-31
feat: Add error event to telemetry

jonathonherbert:
## What does this change?

Adds an error event to the telemetry adapter, and triggers it when there is an error fetching matches. Also adds the URL called to the error message.

## How to test

In the sandbox, trigger an error fetching, perhaps by expiring a cookie or disabling your network connection in your browser developer tools. After a short delay, you should see a network call go out to the event service. It should contain an event that looks a bit like:

```json
[
  {
    "type": "TYPERIGHTER_ERROR",
    "app": "prosemirror-typerighter",
    "stage": "DEV",
    "tags": {
      "message": "https://checker.typerighter.code.dev-gutools.co.uk/checkStream responded with 401"
    },
    "eventTime": "2022-11-01T08:47:19.195Z"
  }
]
```

Better visibility for developers when prosemirror-typerighter has trouble talking to the Typerighter server.

### DIFF
--- a/src/ts/services/adapters/TyperighterAdapter.ts
+++ b/src/ts/services/adapters/TyperighterAdapter.ts
@@ -69,7 +69,7 @@ class TyperighterAdapter implements IMatcherAdapter {
           return onRequestError({
             requestId,
             blockId: input.id,
-            message: `${response.status}: ${response.statusText}`,
+            message: `${response.url} responded with ${response.status} ${response.statusText}`,
             categoryIds,
             type: "AUTH_ERROR"
           });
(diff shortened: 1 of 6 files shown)


src/ts/services/adapters/TyperighterAdapter.ts
```diff
@@ -69,7 +69,7 @@ class TyperighterAdapter implements IMatcherAdapter {
           return onRequestError({
             requestId,
             blockId: input.id,
-            message: `${response.status}: ${response.statusText}`,
+            message: `${response.url} responded with ${response.status} ${response.statusText}`,
             categoryIds,
             type: "AUTH_ERROR"
```
rebecca-thompson, line 74: Non-blocking, but it might be easier when graphing results to have this be the same type as the other errors? ie, `TYPERIGHTER_TELEMETRY_TYPE.TYPERIGHTER_ERROR`

rebecca-thompson (approved):
Have run this locally and it works as expected

PR #245, opened 2022-11-07
fix: Add spellcheck=false to match decorations

jonathonherbert:
## What does this change?

Disable browser spellchecking in matched text.

|Before|After|
|--|--|
|Screenshot 2022-11-07 at 09 31 56|Screenshot 2022-11-07 at 09 32 30|

## How to test

Running locally, and making sure your browser's spellcheck is on, introduce some words to the sandbox that result in a hit from the spellcheck and Typerighter simultaneously ('Keir Starmer' does the trick for the basic spellcheck in Chrome.) They should match the result above.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -66,7 +66,7 @@ const {
     typerighterTelemetryAdapter.matchDecorationClicked(match, document.URL),
   requestMatchesOnDocModified: true,
   adapter: new TyperighterChunkedAdapter(
-    "https://checker.typerighter.local.dev-gutools.co.uk"
+    "https://checker.typerighter.code.dev-gutools.co.uk"
   ),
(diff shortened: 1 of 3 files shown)


pages/index.ts
```diff
@@ -66,7 +66,7 @@ const {
     typerighterTelemetryAdapter.matchDecorationClicked(match, document.URL),
   requestMatchesOnDocModified: true,
   adapter: new TyperighterChunkedAdapter(
-    "https://checker.typerighter.local.dev-gutools.co.uk"
```
jonathonherbert, line 69: Not strictly related but added as part of this PR – it's much more likely developers are working with the CODE sandbox when developing PM-TR locally, and I'm forever correcting when making PRs!

alinaboghiu (approved):

PR #253, opened 2023-08-22
feat: Add additional exports to library so that Composer can use own sidebar

rhystmills:
## What does this change?
Composer will now have it's own copy of the Typerighter sidebar so that UI changes can be iterated upon more quickly. 

This adds extra exports to make that possible.

## How to test
You don't need to do much other than make sure that this repo can build, following the instructions in the readme.


samanthagottlieb (approved):
👍

PR #254, opened 2023-08-22
feat!: Remove sidebar once Composer uses its own.

rhystmills:
This PR removes sidebar code that has been moved to Composer.

Test locally with Composer using yalc, as specified in the readme. Does Typerighter function as expected in Composer?


simonbyford:
Looks good! 

I think you'll need to remove `Sidebar.spec.tsx`. If this spec has any value, I wonder if we should port it to `flexible-content`?

We'll also need to remove the call to `createSidebarView` from `pages/index.ts` (which seems to power https://guardian.github.io/prosemirror-typerighter/)

PR #255, opened 2023-08-23
fix: Fix continuous deployment

rhystmills:
Fixes CD by adding now-required permissions.

As seen in https://github.com/guardian/cdk/pull/1874


rhystmills:
It enabled CD to comment on PRs according to this one: https://github.com/guardian/cdk/pull/1898

PR #257, opened 2023-08-30
feat!: Pass categories through to backend

rhystmills:
## What does this change?
This makes sure that categories passed from Composer are included in the match request to the Typerighter Checker service, where they can be used to filter matches,

## How to test
Test locally with [this Composer branch](https://github.com/guardian/flexible-content/pull/4391) via `yalc` (I can pair to help with this, it's a bit fiddly) - also run this branch of [Typerighter](https://github.com/guardian/typerighter/pull/413) Checker locally.

Good things to check:
- Does Typerighter work as expected?
- If no categories are provided, are all matches returned, regardless of category? (This is the intended behaviour).
- If a category is provided, are only matches with that category returned?

### DIFF
--- a/src/ts/services/adapters/TyperighterChunkedAdapter.ts
+++ b/src/ts/services/adapters/TyperighterChunkedAdapter.ts
@@ -37,7 +37,8 @@ class TyperighterChunkedAdapter extends TyperighterAdapter
       }),
       body: JSON.stringify({
         requestId,
-        blocks
+        blocks,
+        categoryIds
       })
     });
 
--- a/src/ts/state/actions.ts
+++ b/src/ts/state/actions.ts
@@ -41,10 +41,11 @@ export type ActionRequestMatchesForDirtyRanges = ReturnType<
 
 export const requestMatchesForDocument = (
   requestId: string,
-  categoryIds: string[]
+  categoryIds: string[],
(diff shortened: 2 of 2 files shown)


rhystmills:
> I haven't used `yalc` before, what do you mean by test via `yalc`? Happy to pair on this if easier!

`yalc` is a tool that allows you to use your local repo of an npm package instead of using the remote one, it's very handy for testing packages in their consumer before releasing them. In this case we can test this version of `prosemirror-typerighter` in Composer without releasing it. I'm more than happy to pair!

samanthagottlieb (approved):
Tested locally with `yalc` and this works as expected!

src/ts/services/adapters/TyperighterChunkedAdapter.ts
```diff
@@ -37,7 +37,8 @@ class TyperighterChunkedAdapter extends TyperighterAdapter
       }),
       body: JSON.stringify({
         requestId,
-        blocks
+        blocks,
+        categoryIds
```
simonbyford, line 41: It looks like we're sending `categoryIds` twice, once at the root level, once inside `blocks`, was that intended?

Screenshot 2023-09-05 at 15 00 41

PR #259, opened 2023-09-06
feat: Allow exclude list of categories

rhystmills:
## What does this change?
Allow an exclude list of categories rather than allowing only an allowlist.

This means that, in order to an exclude an existing category (i.e. the dictionary category) from a consumer (i.e. Composer), we don't have to fetch the categories, filter it out, then add them to a `MatcherService` - which also means we can initialised the `prosemirrorTyperighterPlugin`, in order to apply the exclusions to the first request (preventing a race condition where we initialised the plugin, get the matcher service, then request a category change).

Dependant on a change in `typerighter`.

## How to test

1. Run Typerighter locally using this branch: https://github.com/guardian/typerighter/pull/427
2. Build `prosemirror-typerighter` locally using this branch and use in `flexible-content` with `yalc` (`npm run yalc` in `prosemirror-typerighter` then `yalc add @guardian/prosemirror-typerighter` in `flexible-content`.
3. Run Composer using this branch: https://github.com/guardian/flexible-content/pull/4391
4. Activate the feature switch via <kbd>shift-f12</kbd> and refresh Composer
5. Try the toggle in both the liveblog and article views. Do matches refresh on enabling or disabling the dictionary, so that only the appopriate matches show?
6. If Collins is disabled, and you de-active the feature switch, does it remain disabled?


src/ts/createTyperighterPlugin.ts
```diff
@@ -108,6 +108,11 @@ export interface IPluginOptions extends PluginOptionsFromConfig {
   adapter: IMatcherAdapter,
 
   typerighterEnabled?: boolean
+  /**
+   * A list of categoryIds to exclude from checks. These can be 
+   * modified on the MatcherService instance.
```
jonathonherbert, line 113:
```suggestion
   * modified on the MatcherService instance if they need to be changed after 
   * the plugin initialises.
```

src/ts/services/MatcherService.ts
```diff
@@ -134,6 +139,29 @@ class MatcherService {
     );
   }
 
+  /**
+   * Request a fetch for matches. If we already have a request in flight,
+   * defer it until the next throttle window.
+   */
```
jonathonherbert, line 145:
```suggestion
  /**
   * Request a fetch for matches as with `requestFetchMatches`, but for the entire document.
   */
```

jonathonherbert (approved):

PR #260, opened 2023-09-08
fix: Show multiple suggestions for dictionary rules

samanthagottlieb:
## What does this change?

This PR makes sure that more than just the first suggested replacement word is shown for Collins dictionary rules. The first suggestion will still appear when users hover over an underlined word. When there is more than one suggestion, a 'See more suggestions' button will appear, which users can click to see additional suggested words.

Before this, only the first dictionary rule suggestion was being shown.

## How to test

This can be tested locally using [`yalc`](https://github.com/guardian/prosemirror-typerighter#testing-locally-in-applications-that-use-prosemirror-typerighter) to locally publish this branch's version of prosemirror-typerighter and to run this package in Composer.

Steps to test:
- create an article
- add a Collins dictionary misspelled word e.g. 'primery' - after running Typerighter you should see a single suggested replacement word with a 'See more suggestions' button underneath - clicking this should reveal alternative replacement words
- add a Style Guide misspelled word e.g. 'learnt' - after running Typerighter you should still see a suggested replacement word

### Before
Screenshot 2023-09-08 at 16 10 58

### After
Screenshot 2023-09-08 at 16 07 08

Screenshot 2023-09-08 at 16 07 19

### DIFF
--- a/src/ts/components/Match.tsx
+++ b/src/ts/components/Match.tsx
@@ -47,8 +47,8 @@ class Match<TMatch extends IMatch> extends Component<IMatchProps<TMatch>> {
       ruleId
     };
 
-    // render suggestions if they exist (the case for dictionary rules), otherwise render the replacement (as sometimes exists for classic Typerighter rules)
-    const suggestionsToRender = suggestions && suggestions.length > 0 ? suggestions : replacement ? [replacement] : [];
+    // render up to 6 suggestions if they exist (e.g. dictionary rules), otherwise render the replacement (as sometimes exists for classic Typerighter rules)
+    const suggestionsToRender = suggestions && suggestions.length > 0 ? suggestions.slice(0, 6) : replacement ? [replacement] : [];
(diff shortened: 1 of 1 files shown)


samanthagottlieb:
> Looking good! Have tested, I've now realised that we sometimes get a very large number of suggestions. Should we cap at 10?
> 
> image
> 
> Would be getting a bit complicated for a ternary at that point, could pass suggestions and replacement to a separate function and handle the logic there.

Good catch on the long list of suggestions! I tried limiting to 10, but it was still fairly length and got cut off at the bottom of the page, so have limited to 6:
Screenshot 2023-09-14 at 13 18 52

It still sometimes get cut off at the bottom, depending on where the word is on the page, which is probably something we should look at fixing in a later PR, e.g.:
Screenshot 2023-09-14 at 13 21 40

rhystmills (approved):

rhystmills:
Works great!

PR #262, opened 2023-09-18
fix: Do not throw an error when a command is dispatched and the document has been destroyed

jonathonherbert:
## What does this change?

Fixes a bug where Prosemirror will throw an error if commands are dispatched after the editor is destroyed.

This PR solves the immediate problem, but it'd be nice to cancel the relevant fetches when the document is destroyed to avoid redundant work.

## How to test

The automated test for this case should pass. We should also see the related [Sentry errors](https://the-guardian.sentry.io/issues/3686433779/?project=26737&query=&referrer=project-issue-stream) drop off once Composer uses this release.

### DIFF
--- a/src/ts/commands.ts
+++ b/src/ts/commands.ts
@@ -436,23 +436,31 @@ export const createBoundCommands = (
   view: EditorView,
   telemetryAdapter?: TyperighterTelemetryAdapter
 ) => {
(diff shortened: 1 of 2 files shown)


samanthagottlieb (approved):
LGTM 👌

PR #264, opened 2023-11-20
feat: Add match feedback inline

jonathonherbert:
## What does this change?

Adds a little form for inline feedback, replacing the Google form in that context. (The Google form remains for general tool feedback in the sidebar, for now.)

feedback

## How to test

Have a play. Does the form work as expected? Any styling or messaging we could improve?

More feedback from users, as a result of a streamlined feedback experience.

### DIFF
--- a/pages/index.ts
+++ b/pages/index.ts
@@ -113,7 +113,6 @@ if (editorElement && sidebarNode) {
     store,
     commands,
     overlayNode,
-    feedbackHref: "http://a-form-for-example.com",
     onMarkCorrect: match => console.info("Match ignored!", match),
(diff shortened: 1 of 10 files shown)


rhystmills (commented):
Much better than our existing feedback form - it would be nice to signpost the anchor text distinctly but no need to do that in this PR. 

I'm finding when I click on "Other" in the prosemirror-typerighter demo that the modal is closing. Code looks great!

rhystmills (approved):
UI working nicely locally in Composer - can't see any requests going out on submit but I'm guessing that's due to config.

Big improvement!

PR #266, opened 2023-11-30
feat!: Handle multi-range matches for ignored ranges.

jonathonherbert:
## What does this change?

Adds internal support for multi-range matches for ignored ranges. Matches can now span an ignored range. Suggestions applied to that range will map around the ignored range, preserving styling where possible.

styling

Ignored ranges are not passed on to the Typerighter service, so this change only affects plugin consumers. For them, this is a breaking change, as some public apis that expect a single `IRange` now receive `IRange[]`.

### How does suggestion insertion work?

If the match is split into multiple ranges, the code attempts to preserve a reasonable split between ranges by allocating the suggestion characters across the original ranges. Extra characters are appended.

For example, given:
  - the match `ex-a-mple`, where dashes denote splits in the match
  - the suggestion `ample`

We get the result `am-p-le`, preserving the position of the ignored ranges, rather than something like `ample--`.

### How does suggestion styling work?

When inserting new content, we map the positions from the insertion back to the original document, and use those positions and the original document to derive marks.

For example, the following text:

```
<strong>exampl</strong> suggestion
        |    |
        1    5
```

might produce the suggestion `example`, spanning `1,6`. This produces the patch `insert e` at `6`. We use the inverse of the transaction [mapping](https://prosemirror.net/docs/ref/#transform.Transform.mapping) produced by the edit to map the position`6` to the original position `5`, from which we can copy the new marks.

The mappings account for ignored ranges, so we only derive styling from unignored copy. See the GIF for examples of this behaviour in practice.

(In adding more tests for this method, I've found that this catches an edge case in our [previous code](https://github.com/guardian/prosemirror-typerighter/pull/127) where we were ignoring styling in single width ranges.)

## How to test

- The automated tests should pass, testing the new functionality.
- Have a play in the sandbox environment – Typerighter now ignores any text that is noted. Add the 'note' mark to text by selecting a range and hitting F10. You should be able to see suggestions being applied correctly, with styling preserved:

## Todo

- [x] ~Test in a consuming application.~ Tested in Composer.

### DIFF
--- a/src/ts/commands.ts
+++ b/src/ts/commands.ts
@@ -31,7 +31,8 @@ import { EditorView } from "prosemirror-view";
 import { compact } from "./utils/array";
 import {
   getPatchesFromReplacementText,
-  applyPatchToTransaction
+  applyPatchToTransaction,
+  getFirstMatchingChar
 } from "./utils/prosemirror";
 import { getState } from "./utils/plugin";
 import TyperighterTelemetryAdapter from "./services/TyperighterTelemetryAdapter";
@@ -361,6 +362,10 @@ const maybeApplySuggestions = (
 
     let textCursor = 0;
 
+    // Ensure the replacement always begins at the first point that the old and new suggestions match.
+    const strToReplace = match.ranges.map(range => tr.doc.textBetween(range.from, range.to)).join('')
+    tr.deleteRange(match.from, match.from + getFirstMatchingChar(strToReplace, text));
+
     // Apply the suggestion to the matched range.
     //
     // If the match is split into multiple ranges, attempts to preserve a reasonable split between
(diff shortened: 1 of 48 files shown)


src/ts/commands.ts
```diff
@@ -355,23 +355,46 @@ const maybeApplySuggestions = (
...
+    // Apply the suggestion to the matched range.
+    //
+    // If the match is split into multiple ranges, attempts to preserve a reasonable split between
+    // ranges by allocating the suggestion the same number of characters as the original range. Extra
+    // characters are append.
+    //
+    // For example, given:
+    //   - the match 'ex-a-mple', where dashes denote splits in the match
+    //   - the suggestion 'ample'
+    // We get the result 'am-p-le'.
```
rhystmills, line 381: Bit confused by this, should the result by a-mple ?

jonathonherbert: It's perhaps an unfortunate example: we keep the splits where they are, and change the characters. Not clever, does the job: in this case,

```
ex-a-mple
123456789
  | |
  ^ ^

becomes

am-p-le
1234567
  | |
  ^ ^
```

This could perhaps be improved by doing something more clever: like looking at the diff to see where the strings first begin to match, and counting from there.

rhystmills: Ah I see - the existing solution seems like a decent enough solution to me.

rhystmills (approved):
Amazing feature and should help users a lot in Composer, especially with the notorious missing/double spaces. Working when testing locally and in the tests, code looks good.

For readers - there will need to be a few (not too intensive) tweaks in Composer before this works there.

PR #279, opened 2025-01-13
fix: Only update consumers when state changes

jonathonherbert:
## What does this change?

Doing some performance profiling in the Guardian's CMS, Composer, it looks like we’re spending a lot of time rendering Typerighter’s UI, even when it’s not onscreen and there are no matches to display.

This is because the Typerighter plugin store is emitting events on every state update, even when nothing's changed.

Rather than burden the consumer with this check, this PR solves the problem at source for empty documents. When the plugin view updates, we now check to see if the plugin state has changed before updating store subscribers. This helps reduce unnecessary rerenders when subscribers are rendering UI in response to state changes.

## How do we check the plugin state?

The best way would be to run a reference check: `prevState === state`. Unfortunately, we always update dirty ranges when the document changes, although listeners never consume them.

This PR makes the state we share with consumers a subset of the plugin state, removing `dirtiedRanges`. Because we must create a new object identity when we immutably extract the subset of the state we need, we must use shallow equality on the resulting object to determine whether the state should be broadcast to consumers.

I've added [immer](https://immerjs.github.io/immer/) to make updating the state immutably easy, and vendored a shallow equality utility in [8dfc0c6](https://github.com/guardian/prosemirror-typerighter/pull/279/commits/8dfc0c6032f629b4c940bd34be253e4ba718860e). (I implemented an immutable update that took care to avoid reference swaps in [bda61e7](https://github.com/guardian/prosemirror-typerighter/pull/279/commits/bda61e7d976ff8e063263f69d1eb45c95460734e), but I think [3e9f553](https://github.com/guardian/prosemirror-typerighter/pull/279/commits/3e9f5539348b985272ba189c71607cd6811f9ece) is cleaner, and opens the door to making additional optimisations of this sort more readable if they're necessary.)

This PR does not currently optimise for state changes when we're processing changes as a result of dispatching reducer actions; this PR simply optimises the case when Typerighter is doing nothing as the document changes.

## How to test

- The automated tests cover the state identity change and the subscription check, for transactions that affect the document, and transactions that don't.
- Here's some profiling in our CMS: the scripting cost of 32 keystrokes in a large document, tested on Chrome 131.0.6778.205 a 14-inch Macbook Pro, M1 Max, 2021:

Iteration|1|2|3
-- | -- | -- | --
Before, ms | 2080 | 2116 | 2083
After, ms | 1465 | 1428 | 1473

Improved performance in editors that use this plugin.

Works towards https://github.com/guardian/flexible-content/issues/5169

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@mui/material": "^5.9.0",
         "@popperjs/core": "^2.4.4",
         "diff": "^4.0.2",
+        "immer": "^10.1.1",
         "lodash": "^4.17.11",
         "prosemirror-changeset": "^2.2.0",
         "prosemirror-example-setup": "^1.2.1",
@@ -4999,6 +5000,15 @@
       "dev": true,
       "engines": {
         "node": ">= 4"
+      }
+    },
+    "node_modules/immer": {
+      "version": "10.1.1",
+      "resolved": "https://registry.npmjs.org/immer/-/immer-10.1.1.tgz",
+      "integrity": "sha512-s2MPrmjovJcoMaHtx6K11Ra7oD05NT97w1IC5zpMkT6Atjr7H8LjaDd81iIxUYpMKSRRNMJE703M1Fhr/TctHw==",
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/immer"
       }
     },
     "node_modules/immutable": {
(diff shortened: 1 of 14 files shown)


KaliedaRik (approved):
Confirm that I can run the PR branch in my LOCAL environment, and all tests pass when running `npm run test` 

Happy to approve.

PR #300, opened 2026-09-16
fix: guard tooltip offset calculation against empty/stale getClientRects()

davidfurey:
> [!NOTE]
> This PR was opened by Claude Sonnet 5 based on an interactive session and while I have reviewed the output, I am not very familiar with this codebase

## Problem

Reported via Sentry as [guardian/flexible-content#6722](https://github.com/guardian/flexible-content/issues/6722) (COMPOSER-64Q, duplicate COMPOSER-88T): a `TypeError: Cannot read properties of undefined (reading 'left')` thrown from inside Popper.js while positioning the `MatchOverlay` hover tooltip. 343 events in the last 30 days (18,120 lifetime), first seen 2022-07-28.

## Root cause

`MatchOverlay`'s Popper `offset` modifier (`getOffsets`) reads `.left`/`.height` off the first and last rects returned by the hovered match decoration's `getClientRects()`, without checking that they exist:

```ts
const rects = referenceElement?.getClientRects();
const hoverRect = rects[currentRectIndex];
const lastRect = rects[rects.length - 1];
const x = hoverRect.left - lastRect.left; // throws if hoverRect/lastRect are undefined
```

If the decoration's range/DOM node is removed from the document while it's still hovered (e.g. the underlying text is edited, or the block unmounts) before Popper's async offset calculation runs, `getClientRects()` can return an empty list, and `hoverRect`/`lastRect` are `undefined`.

## Fix

Bail out to the default offset (`[0, yOffset]`) when the hover/last rect is missing, instead of crashing.

## Test

Added a regression test in `MatchOverlay.spec.tsx` that renders the overlay, hovers a match whose decoration element has no layout (as is always the case in JSDOM, and matches the "stale/removed" condition in production), and asserts no error is thrown. It reliably reproduces the exact reported stack trace (`MatchOverlay.tsx` → `@popperjs/core/offset.js` → `createPopper.js#forceUpdate`) before the fix, and passes after it.

`npm test` (177 tests) and `tsc --noEmit` are green.


jonathonherbert (approved):
Test is red w/o fix, and then goes green. Explanation checks out. 👍

PR #301, opened 2026-09-18
feat: Make feedback mode configurable via restored feedbackHref param

jonathonherbert:
## What does this change?

Make feedback mode configurable via restored feedbackHref param. At the moment, the inline feedback mode does not have an adequate integration downstream, and so it blocks adoption. This PR makes it optional.

## How to test

Import via yalc or similar (see README.md) in Composer. Click on the feedback link in the match popover. It should link to a Google Form, as before.

- [x] Tested locally in Composer, works as expected


waisingyiu (approved):
It looks great!

